PR #1, opened 2021-07-16
First ADR submitted around the database bottleneck

jimleroyer:
# Summary | Résumé

It is incomplete and a draft at that moment. Up for review though to complete with the team.

# Test instructions | Instructions pour tester la modification

Read it!

# Help requested | Aide requise

Yes.

      langues officielles?
      fonctionnalité existante?
- [ ] Does this introduce any security concerns? | Est-ce que ça introduit des
      préoccupations liées à la sécurité?


records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Insert the notifications as a batch
+
+As we streamline the notification save with the previous option and got
+these under Celery land, we can now batch these easily. There is a 
+[celery-batch](https://pypi.org/project/celery-batches/) project which 
+exists that might offer us the option to easily achieve this with minimal
+setup.
+
+### Tweaking autovacuum options
```
wmoussa-cds, line 89: I feel anything that we would try to control or manage a built in database mechanism of cleaning is a risky move, and a temporary solution at best. I'd prefer we look at reviewing the architecture of the solution.

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,135 @@
...
+type, scheduling and used key.
+4. In the case of an email that was requested through the API, the email
+will be sent to the AWS SES provider (the only supported at the time of
+this writing).
+5. As multiples emails are sent and some receive a response receipt, multiple
+updates on the database are performed.
+6. In the meantime, numerous count operations are performed to track the
+current delivered emails and various metrics.
+7. As there are more and more commits performed on the database, for each and
+every notification, the database triggers an autovacuum action to clean up the
```
wmoussa-cds, line 46: does the autovacuum slow dow the whole database or just activities on one table?

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,135 @@
...
+* the AWS instance classes,
+* the resources requests and limits of Kubernetes containers,
+* the horizontal pod autoscaling.
+
+These serve well the application but there were already signs of the
+application's limit: we rarely if ever meet our SLAs and we had regular daily
+incidents that foretold of the difficulties ahead.
+
+This document propose various options to address thorny issues that our
+database is having.
```
wmoussa-cds, line 29: where can we find the current database diagram? a link in this ADR will give more context

jimleroyer: PR is incomplete and diagrams to come soon.

records/2021-07-16.scalability.database-bottleneck.md
jimleroyer: It slows down all components interacting with the database, so I infer the whole database for now.

records/2021-07-16.scalability.database-bottleneck.md
jimleroyer: This is our best mitigation for short term, as we are losing notifications. A combination of manual run (but keeping automatic runs) + options tweaking could be beneficial. I am working on recommendations from the AWS expert on this. We also now have a quasi similar production-like environment in staging that we can use to test the configurations first on knowing the risk of these while [we run load/stress tests in these](https://github.com/cds-snc/notification-api/tree/master/tests-perf/locust).

Although, indeed, I agree that solving the design is the long term solution, to me at least, but I wouldn't exclude a more proper tweaking of these to accommodate a burst profile like Notify has on long-term with careful grooming in staging.

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,135 @@
...
+### Database connections pooling
+
+To maximise the distribution of database connection, we can have a
+pooler that would sit in-between the Notify stack and the AWS RDS
+Aurora Postgres instances.
+
+RDS Proxies would exists for this purpose. Apparently, we could create an
+endpoint each for both readers and the writer. This could co-exist nicely
+with our existing application logic which dispatch database sessions to
+either the readers or writer.
```
CalvinRodo, line 125: This seems like it should be the easiest solution and **should** have no impact on your application.

jimleroyer: Yep; we should do it no matter what.

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,187 @@
...
+
+Upon receiving a notification from the API, do not save it in the current
+*notification-api* process but defer it to a celery task. This does not
+address the root cause of our issue but offer many benefits, namely:
+
+1. This is an iterative improvement where we send the database save into
+   [Celery land](https://docs.celeryproject.org/en/stable/). Having these into
+   a Celery tasks means we can more easily control the flow of these operations
+   and control these with all of the Celery features. Specifically, we can batch
+   the notifications using [celery-batch](https://pypi.org/project/celery-batches/).
```
dorianjp, line 99: Moving db heavy work to async celery makes good sense, but do we want to add dependency complexity to overcome code complexity? I see this project was experimental and may not batch anything db related as hoped, at least at first glance of the code and readme.

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,187 @@
...
+Because our readiness probe reports that our pods are not ready anymore and
+hence, get detached from the load balancer, which **then cause notifications to
+be lost forever**, we could tweak it on the short term to accommodate our current
+situation and stop the bleeding.
+
+Normally, a delayed database should not be normal. We know in our current
+situation though that the database is still up and running. We'll receive paging
+alerts anyway if it goes down. We should definitely put back the database
+check once the design is fixed.
+
```
dorianjp, line 170: It's a good call to ease on the readiness check as you're aware a slow response doesn't mean a dead service.

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,187 @@
...
+### Notifications table sharding
+
+_To be completed_.
+
+## Decision
+
+Our current recommendations at the time of this writing:
+
+1. Stream the notifications save from the API to a Celery task.
+2. Insert the notifications as a batch once #1 is done.
```
dorianjp, line 180: Are there plans to investigate the heavy queries issue on the DB when the celery task has additional work to do?

records/2021-07-16.scalability.database-bottleneck.md
dorianjp: Don't underestimate just provisioning larger instances to buy time. I am not a big fan of buying a way out of an efficiency problem, but I believe the cost difference from existing to r6large or r6xlarge is minimal when compared to costly developer time. It could be considered kicking the can down the road, but with a solid plan to mitigate and reduce unnecessary load, it can be a worthwhile investment.

dorianjp (commented):
Great write-up @jimleroyer.

I am curious what the thoughts are from the team (@willeybryan , @ClementineHahn) on notify performance is overall in regards to handling unpredictable bursty traffic. I mean every service on the planet has a point where enough traffic will knock it out, but what is an acceptable buffer before that's the case? 1 unexpected burst, 2x, 5x, or 10x?

willeybryan:
I agree completely @dorianjp I too hate the idea of buying our way out of an efficiency problem. If feels like we have taken the easy way out. In the interest of full disclosure this was my Idea. With the performance problems escalating and Antoine's departure I made the call and got approval from Elodie to throw more money at the problem to tide us over. Sadly We have been dropping a few messages and that needs to stop ASAP.

GC Departments are dead set on sending email blasts for a wide variety of things as it has been practice for decades and they architect services around it. We need to implement a better solution for handling bursty traffic before we start dropping important ones. While more people are doing it on demand, there are a number of use cases the GC isn't going to give up on where message blasts are the preferred mechanism. We recently enabled the API call to take CSV data/multiple JSON entries in one call and implemented a bulk API we can throttle down or put layers in front of. 

My hope is that we can buy time to fix the problem properly, build proper load testing, and better scale Notify with the traffic we will get in the future.

records/2021-07-16.scalability.database-bottleneck.md
jimleroyer: No plan for now. It's worth checking, but not where the database spend most of its time. The AWS performance insights dashboard indicates that commits cost the most.

image

dorianjp:
> I agree completely @dorianjp I too hate the idea of buying our way out of an efficiency problem. If feels like we have taken the easy way out. In the interest of full disclosure this was my Idea. With the performance problems escalating and Antoine's departure I made the call and got approval from Elodie to throw more money at the problem to tide us over. Sadly We have been dropping a few messages and that needs to stop ASAP.
> 
> GC Departments are dead set on sending email blasts for a wide variety of things as it has been practice for decades and they architect services around it. We need to implement a better solution for handling bursty traffic before we start dropping important ones. While more people are doing it on demand, there are a number of use cases the GC isn't going to give up on where message blasts are the preferred mechanism. We recently enabled the API call to take CSV data/multiple JSON entries in one call and implemented a bulk API we can throttle down or put layers in front of.
> 
> My hope is that we can buy time to fix the problem properly, build proper load testing, and better scale Notify with the traffic we will get in the future.
> 
> I don't know how many is X so I don't know what a good multiple is.

It was a bit of a leading question @willeybryan, but as it stands, a 2x burst is enough to take portions of the service down intermittently, and we should be prepared for factors much larger than 2x at any time as we don't get advance notice of heavy usage from clients. It's great to tune for efficiency where we can, but it's worthwhile investigating design patterns to completely alleviate our burst bottleneck entirely. At the end of the day, 50k-200k requests on a single API endpoint within 15-60 minutes is something we should be able to handle without notify breaking a sweat.

records/2021-07-16.scalability.database-bottleneck.md
jimleroyer: I'd say let's focus on the goal and not the means: I'd have anything that can batch our calls at the moment. 

Regarding dependencies: granted, we should be careful in to import experimental code in the project.

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,187 @@
...
+database is having. [Click here for schema of our database](diagrams/2021-07-16.scalability.database-bottleneck/relationships.real.compact.png)
+at the time of this writing.
+
+## Timeline of an incident
+
+1. A blast of notifications is sent to Notify through the API.
+2. As the notification-api component saves the notifications to be sent in
+   the database, it commits each of these individually[1].
+3. The notification is then sent to the appropriate queue, depending on its
+   type, scheduling and used key.
```
Moro-Code, line 38: There is a missing step here. Once the notification goes to the queue the status of the Notification is updated from created to sending. This is important because its an update operation ( under load can trigger autovac )

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,187 @@
...
+   to stop receiving any request. It affects all instances of the API and Admin
+   which both heavily relies on calls to the same database.
+10. As the database slowed down to answer the API, the latter does not stop
+   receiving and processing notifications. It continues to makes calls to the database
+   and consumes more database connections as previous ones are still busy.
+11. Errors of maximum number of database connections getting reached start
+   to appear at that point.
+12. The API instances are down as these are detached from the load balancers,
+   the database is overloaded with connections and can't take any more. **We are
+   losing notifications** because we don't save these when the API is unavailable.
```
Moro-Code, line 65: This is super clear and detailed Jimmy. It perfectly sums up what happens

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,187 @@
...
+of high traffic, it becomes a performance issue. An apt analogy is possible
+with someone getting to the cashier in a grocery store to pay their items:
+imagine that you would have to pay for each these items individually, i.e.
+a) scan one and only one item, b) ask for the total, c) pay with cash or card,
+d) repeat for all remaining items. This will take quite a long time compared
+to paying for all items in one batch within one transaction. Saving one
+notification every time in the database is akin to paying for one grocery item
+individually in one transaction, except you have potentially thousands of items
+remaining afterward. Ideally, we want to buy/save all for all of these in one
+go.
```
Moro-Code, line 78: I like this example ;)

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,187 @@
...
+Upon receiving a notification from the API, do not save it in the current
+*notification-api* process but defer it to a celery task. This does not
+address the root cause of our issue but offer many benefits, namely:
+
+1. This is an iterative improvement where we send the database save into
+   [Celery land](https://docs.celeryproject.org/en/stable/). Having these into
+   a Celery tasks means we can more easily control the flow of these operations
+   and control these with all of the Celery features. Specifically, we can batch
+   the notifications using [celery-batch](https://pypi.org/project/celery-batches/).
+2. The number of database connections will be lowered in the API component
```
Moro-Code, line 100: I would add such that even if the database is slowed the API route to create notifications will not be slowed causing the EKS chain reaction of detaching the load balancer

records/2021-07-16.scalability.database-bottleneck.md
Moro-Code: @dorianjp we already have a r6large in prod and staging actually. This is not a problem we're going to easily be able to buy our way out of unfortunately

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,187 @@
...
+
+There are a few autovacuum configuration that we can tweak in order to delay
+the autovacuum action that triggers a performance slowdown of our database:
+
+* **autovacuum_scale_factor** (default: 0.2): we could use this to delay the vacuum, (suggestion 0.3).
+* **cost_delay**: We could use this to delay by increasing the threshold.
+
+### Manual execution of the vacuum process on the `notifications` table
+
+An option is to keep the autovacuum enabled but trigger a manual vacuum prior
```
Moro-Code, line 121: Should mention the con with this. We don't really know when we will get these bursts. We also don't know if the burst will still trigger the autovacc activity

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,187 @@
...
+to traffic time. This would better prepare the database for upcoming load.
+
+Don't run `vacuum FULL`: this will take on most database resources and prevent
+proper operation of the database.
+
+### Disable the autovacuum process on the `notifications` table
```
Moro-Code, line 127: I believe the recommendation was only to do this before the burst to indeed see if autovac is the issue rather than being a recommendation to do this. This was my understanding at least

Moro-Code (commented):
Great doc ! Very clear I love the format

records/2021-07-16.scalability.database-bottleneck.md
jimleroyer: Took it from someone you know. 😅

records/2021-07-16.scalability.database-bottleneck.md
jimleroyer: Yes it's my understanding too, for exploration purpose. The recommendation was to do it 1-2 days from the AWS expert. Let me clarify.

records/2021-07-16.scalability.database-bottleneck.md
```diff
@@ -0,0 +1,187 @@
...
+
+To avoid the autovacuum to process the `notifications` table during the daily
+blast of notifications, we could disable the autovacuum to prevent it from
+triggering. This need to be done carefully, as a maximum limit of 4 billion
+transactions can occur in PostgreSQL. The autovacuum cleanup reset this
+transaction number. If this threshold is reached, our database will be severely
+affected.
+
+This is a risky move and certainly not a long/mid term solution.
```
sastels, line 136: Is this actually risky? How long could we go without any vacuuming before things fell over? Months? Could we just have some sort of manual vacuum overnight (or, alternatively, can we set autovacuum so that it doesn't run during business hours?)

records/2021-07-16.scalability.database-bottleneck.md
jimleroyer: I can expand on this, mentioning that we need to identify the bursts. 

We do tend to know when these happen, as it's 95% of the time around 13h-15 EST. We cannot assume that will always be the case.

I personally do think that bursts will trigger autovacuum, i,e. the manual vacuums might relieve the autovacuum triggered by bursts, but I don't put much hope on it. We'd need to test and see, could be going against our expectations. 😀

records/2021-07-16.scalability.database-bottleneck.md
jimleroyer: Good point, although the API could still get detached from the load balancer if the database get slowed down, unless we do the suggestion for *Tweak the readiness and liveness Kubernetes probe*. But the API will be offloaded of DB operations and should process faster.

records/2021-07-16.scalability.database-bottleneck.md
jimleroyer: Hmm this is covered by step #5 but maybe the chronology isn't clear here. Let me rework this.

jimleroyer:
@willeybryan @dorianjp The proposed changes here are to address some design issues; this is a [goal for adopting ADRs](https://adr.github.io/). It's the first of many, incremental. I think the team is a lot more aware of the issues now.

Ultimately, I hope we can streamline the ADR process and bring valuable technical/business feedback from the team and the dev community, with the goals to be transparent and vocal about these. There will be more ADRs soon. **Anyone** should be able to propose these. Ideally, we'll concert and plan accordingly. Given we're not short staff and able to.

Regarding how much burst we can expect: I think these changes in this ADR will allow us to go way higher. Our daily average is 36,778 notifications; our daily record is 280,147 notifications under a 2 hours period (which did lead to an incident): that produces a 7X multiplier. We may be able to replay that record day soon with the proposed changes in this ADR with much better results.

> but as it stands, a 2x burst is enough to take portions of the service down

More like 3X burst, it's when we get past the 100,000 notifications that we can get caught. We went through 83,049 notifications 3 days ago and had no incidents. The database is the bottleneck. We fix this, we'll get to more interesting problems to fix, design wise IMO. And we'll be able to leverage our recent work on load/stress tests and *prod-like* staging environment to increase our proactive capacity.

records/2021-07-16.scalability.database-bottleneck.md
jimleroyer: It's hard to tell for me without exploration. I am not sure if the transaction ID for postgres would shoot to the roof without constant autovacuum. We could know a bit more by disabling in staging.. but given the common wisdom provided to be careful with this, if we would assume we are safe and something happens, that will be critical.

Moro-Code (commented):
I'm just wondering for the sequence diagram in the QueueEmail. There are no updates that happen. All the updates are done within the celery tasks to my understading

PR #3, opened 2021-08-16
Move the Notify API from Kubernetes to AWS Lambda

smcmurtry:
# Summary | Résumé

Up for discussion.

### DIFF
--- a/records/2021-08-16.scalability.move-api-to-lambda.md
+++ b/records/2021-08-16.scalability.move-api-to-lambda.md
@@ -8,13 +8,34 @@ Date: 2021-08-16
 
 ## Context
 
-The SRE team is encouraging us to run the Notify API as a Lambda function in AWS. 
+The SRE team is encouraging us to move the Notify API from Kubernetes to AWS Lambda. AWS Lambda functions allow code to be run in a highly scalable way - we can have hundreds of lambda functions running simulateously, and this auto-scales to meet the number of requests being made to the API. The benefits of doing this are thought to be the following:
+
+- less complexity for the Notify team to manage, compared to Kubernetes
+- more scalable API
+- less costly compared to kubernetes + Amazon EKS
+
+At the time when Notify was being deployed by CDS, AWS Lambda had a smaller maximum size for a lambda function. Notify was over this limit and so @max deployed it a different way.
+
+Since then, AWS has added support for Docker images and increased the size limit for these to 10 GB. This means it is now possible to deploy the Notify API as a Lambda function.
+
+CDS has now used lambda functions to run the covid alert metrics collection and aggregation (https://github.com/cds-snc/covid-alert-metrics-terraform/tree/main/aws). Lamba functions helped that part of the service to scale well.
+
+## Proposed architecture
+
+AWS Lambda + API Gateway
+
+## Additional considerations
+
+AWS Lambda functions exhibit "cold-start" behaviour, where they can take more than 1 second to respond if they [have not been invoked recently](https://aws.amazon.com/blogs/compute/operating-lambda-performance-optimization-part-1/#:~:text=The%20duration%20of%20a%20cold,test%20functions%20than%20production%20workloads). If we do not mitigate this, the admin site will appear unresponsive at times. One possible mitigation would be to ping the api at some frequency - the heartbeat service we already have in Notify could be used for this.
+
+Another consideration will be the effects on other parts of Notify. If the API can scale much better than in the past, will some other parts of the system become overloaded? The may not be an issue because we queue our upcoming jobs with Celery, but we should do significant load testing before running lambda functions in production.
+
+## Load testing results
+
 
 ## Decision
 
-Our current recommendations at the time of this writing:
-
-1. Run the Notify API as a Lambda function in AWS
+TBD based on load testing results
 
 ## Consequences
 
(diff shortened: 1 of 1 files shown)


records/2021-08-16.scalability.move-api-to-lambda.md
```diff
@@ -0,0 +1,42 @@
...
+
+Since then, AWS has added support for Docker images and increased the size limit for these to 10 GB. This means it is now possible to deploy the Notify API as a Lambda function.
+
+CDS has now used lambda functions to run the covid alert metrics collection and aggregation (https://github.com/cds-snc/covid-alert-metrics-terraform/tree/main/aws). Lamba functions helped that part of the service to scale well.
+
+## Proposed architecture
+
+AWS Lambda + API Gateway
+
+## Additional considerations
```
CalvinRodo, line 27: Might want to consider the tradeoffs of local testing, I'm assuming that Kubernetes might be an easier solution to test locally since as long as you have docker desktop installed you can probably bring up the test system pretty easily. 

Lambdas make local testing a bit more difficult by forcing you to either wire up some docker containers using aws-lambda-rie or use AWS SAM to orchestrate the work.

smcmurtry: Thanks @CalvinRodo I added this in.

@jimleroyer Are you or anyone else doing local testing with Kubernetes?

records/2021-08-16.scalability.move-api-to-lambda.md
```diff
@@ -32,7 +32,15 @@ Another consideration will be the effects on other parts of Notify. If the API c
 
 ## Load testing results
 
+We ran a load test where we simulated 5,500 requests per minute to the api using locust, to see if the new lambda deployment could meet existing peak loads. The these requests were made to the `/v2/notifications/email` endpoint and were split into 2 services to make the test more realistic. Here are the reports:
+- [Report for Service 1](attachments/report_1629229241.7761497.html)
+- [Report for Service 2](attachments/report_1629229258.5504797.html)
 
+There were 0 failures, and median response time was 110ms for both services and 99th percentile response time was 250ms for Service 1 and 230ms for Service 2 over a 7 minute test.
```
dj2, line 41: Do you have these numbers for the k8 version so they can be compared?

records/2021-08-16.scalability.move-api-to-lambda.md
```diff
@@ -32,7 +32,15 @@ Another consideration will be the effects on other parts of Notify. If the API c
...
+
+We had provisioned 2 concurrent lambda functions to always run, and these handled about 10% of the requests, with the other 90% going to lambda functions that were provisioned on-demand.
+
+These results demonstrate that the lambda deployment should be able to hande the current peak loads.
```
dj2: nit: s/hande/handle

records/2021-08-16.scalability.move-api-to-lambda.md
smcmurtry: No, but I will get them!

smcmurtry: Done.

records/2021-08-16.scalability.move-api-to-lambda.md
```diff
@@ -0,0 +1,64 @@
...
+
+Another consideration will be the effects on other parts of Notify. If the API can scale much better than in the past, will some other parts of the system become overloaded? This may not be an issue because we queue our upcoming jobs with Celery, but we should do significant load testing before running lambda functions in production.
+
+Lambdas may make local testing a bit more difficult by forcing us to either wire up some docker containers using aws-lambda-rie or use AWS SAM to orchestrate the work. With Kubernetes, testing locally is easier since docker desktop can be used to bring up the test system.
+
+## Load testing results
+
+We ran a load test where we simulated 5,500 requests per minute to the api using locust, to see if the new lambda deployment could meet existing peak loads. The these requests were made to the `/v2/notifications/email` endpoint and were split into 2 services to make the test more realistic. Here are the reports:
+
```
jimleroyer, line 38: It might be worth describing the pacing time and how bursty the tests were and add how unusual it is from regular API loads in production (hence the failure is "acceptable" because the traffic is out of normal bounds), but yet shows that lambdas can withstand more compared to our current EKS/machines/pods configuration.

jimleroyer (commented):
What would be required to add scheduled provisioning of the lambdas (in usual high traffic time)? Does that require some programming or can we do it through the AWS dashboard/Terraform?

records/2021-08-16.scalability.move-api-to-lambda.md
dj2: I'd also be curious what the load test says without a provisioned lambda? What kind of times do you get with all cold starts? (Or do you expect your traffic to be on the order of 1 request in the time it takes the lambda to shutdown?)

smcmurtry: > I'd also be curious what the load test says without a provisioned lambda? 

We determined from the AWS logs that only about 10% of requests were handled by the provisioned lambdas, and the other 90% were handled by non-provisioned lambdas. So the response time stats were dominated by the responses of non-provisioned lambdas. Therefore I would expect these stats wouldn't change by much if we hadn't provisioned any lambdas. 

This chart is a bit hard to read but it shows the approx. 90%/10% split:
image (2)

smcmurtry:
> What would be required to add scheduled provisioning of the lambdas (in usual high traffic time)? Does that require some programming or can we do it through the AWS dashboard/Terraform?

Looks like we can do this just with AWS/Terraform. I added a comment here: https://github.com/cds-snc/notification-adr/pull/3/commits/8459436738557b258d9557535ae6e61305cb865e#diff-72c06ae1179c4b487b6ea276cc04e09c93c55734cba416b8d3c32177e003ea4eR29

PR #6, opened 2021-10-20
Archive document-download-frontend

sastels:
# Summary | Résumé

ADR regarding archiving `document-download-frontend`.

# Help requested | Aide requise

Please review, comment and suggest.


jimleroyer:
I spoke to Antoine about this and he said he kept the front-end in case that we still needed it, i.e. this might happen in the near future that we will have to bring back a download front-end in order to provide users more information, have confirmation or for more advanced user's verification. In the meantime though, we could totally shut down the front-end to eliminate maintenance effort on it.

yaelberger-commits:
Just needs a review to decide if we want to do this or not

yaelberger-commits:
Decided yes from ADR

jimleroyer (approved):
Let's update the status decision sections and merge it!

PR #15, opened 2022-02-28
Make security assessment checkpoint explicit

wmoussa-cds:
With the Notify ATO approved as of Feb 2022, we should keep an eye on any major system decision that can jeopardize the product's security posture.


jimleroyer (approved):

jimleroyer:
Works with me.

PR #16, opened 2022-03-08
Priority lanes

jimleroyer:
# Summary | Résumé

Current draft for enforcing priority lanes within Notify.

* Possibly if we do not crypto sign the notifications within the new storage locations backing up the priority lanes.

      langues officielles?
      fonctionnalité existante?


records/2022-03-08.priority-lanes.md
```diff
@@ -0,0 +1,32 @@
...
+Database operations are all sent through the database-tasks queue/worker. This has a benefit of lining up and tracking a batch of operations to reduce the load. A major downside is that this is a bottleneck for high priority database operations such as saving a notification that needs to be sent in the next seconds. There are use cases when GCNotify is constantly hit and the database-tasks queue grows bigger. There are no FIFO strategy set on this queue as we are targeting high volume processing. This means though that no priority distinction is possible for operations going through it.
+
+## Options
+
+Listed in this section are different options available to us to improve our priority lanes.
+
+### Improving database-tasks SQS queue
+
+* Making it faster
+* Getting high priority database operations out of it; but these need to be low volume.
```
patheard, line 24: This could be an interesting challenge if people just switch everything over to high priority.  We'll need a way to validate that a service's high priority notifications are a valid use case.

patheard, line 24: This could be an interesting challenge if people just switch everything over to high priority.  We'll need a way to validate that a service's high priority notifications are a valid use case.

jimleroyer: This comment came out as a duplicate, not sure why, but I will reference previous reply I had on the other identical comment over here: https://github.com/cds-snc/notification-adr/pull/16#discussion_r822816672

records/2022-03-08.priority-lanes.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Redis inboxes
+
+We currently have 2 inboxes for batch saving via Redis (email and sms). Batches of notifications are taken out in the order they were POSTed. This means high priority notifications might have to wait behind a large number of bulk notifications.
+
+## Current Implementation
+
+![Current Batch Saving](https://user-images.githubusercontent.com/8869623/160944266-48cf19fb-ad07-411b-8f73-88616697f9a8.jpeg)
+
+We store all notifications that come throught the API into a Redis List (Inbox). A process takes the notifications from Inbox and makes lists of a certain chunk size. Each list get processed as a single unit and gets saved to the Database. Instead of saving a single notification at a time, we are now saving a list of notifications (batch) in a single transacation.
```
jimleroyer, line 21: Hmm while the priority lane will get replicated at the batch level, it's also meant for the dedicated database save. So it's not all tied to the batch saving: we should have done this batch before batch saving implementation. 

The current implementation is currently honored through the various celery tasks and workers but not upstream (db save + batch saving levels). We need to adjust these upstreams.

jimleroyer (commented):
Provided a few comments. We can sync offline on these as it might go faster.

sastels:
Another option would be to just have one lane (basically, what we have now) but only create inflights from priority messages if there are any. If there aren't any, then you can make an inflight with normal or bulk. This would have the advantage of simplicity but the significant drawback with this is that if could stall normal or bulk messages indefinitely (so IMO the other option discussed is preferred).

jimleroyer:
@sastels I see what you mean now. We could not do that with Redis list though as we'd have to cycle through the whole list and that would take a performance hit. We'd have to use a [hset](https://redis.io/commands/hset/) maybe.. and I am not sure the entries can be appended. Good to brainstorm alternatives though, thanks for suggesting. 👍

records/2022-03-08.priority-lanes.md
```diff
@@ -0,0 +1,76 @@
...
+
+Listing contextual elements to consider prior to offer suggestions and deepen our understanding of current limitations on priority lanes in GCNotify.
+
+### Redis queues
+
+In order to buffer up notifications between the API and the celery tasks, GCNotify currently keeps these in two Redis queues: one for SMS and one for emails. In Redis land, batches of notifications are taken out in the order these were received. This means high priority notifications might have to wait behind a large number of bulk notifications.
+
+## Current notification pipeline flow
+
+![Current notification flow](https://raw.githubusercontent.com/cds-snc/notification-adr/main/records/diagrams/2022-03-08.priority-lanes/current-lanes-setup.drawio.png?raw=true "Current notification flow")
```
jimleroyer, line 19: current-lanes-setup drawio

records/2022-03-08.priority-lanes.md
```diff
@@ -0,0 +1,76 @@
...
+
+2. ### Specific Beat Functions
+
+   The current beat functions have a cadence of 10 seconds. We can have beat workers scheduled at a faster cadence for higher priority notifications. For example, bulk can be scheduled at 10 seconds, normal at 5 seconds and high priority every second.
+
+3. ### Improve Queuing
+
+   We currently send all notifications to the database queue. We can split the priority/normal/bulk notifications to be sent to corresponding priority queue. We will have to create SQS queues to and forward the notifications to corresponding priority queue.
+
+![Priority lanes extended](https://raw.githubusercontent.com/cds-snc/notification-adr/main/records/diagrams/2022-03-08.priority-lanes/next-lanes-setup.drawio.png?raw=true "Priority lanes extended")
```
jimleroyer, line 66: next-lanes-setup drawio

jzbahrai (approved):
thanks for adding and updating this! looks good to me.

PR #18, opened 2022-03-23
docs: Lambda API environment variable config

patheard:
# Summary

Document the work that has taken place to sync the Lambda API environment variable configuration with the Kubernetes API deployment.

Rendered content of the change:
[2022-03-23.deployment.lambda-api-environment-variables.md](https://github.com/cds-snc/notification-adr/blob/feat/lambda-api-config-sync/records/2022-03-23.deployment.lambda-api-environment-variables.md)

# Related
* cds-snc/notification-planning#421
* cds-snc/notification-planning#458


patheard:
Would you like links added to the work we did to make this happen?  I can see it being useful for future history, but it also creates an odd scenario where an ADR (in review) already has the proposed work completed.

jimleroyer (approved):
Thanks @patheard for this work. Do you think we need further ADR updates since this was opened > 15 days ago?

patheard:
No, I don't think so.  I'm going to create a new ADR for some upcoming config change proposals.

PR #20, opened 2022-04-14
Explain secrets usage within Notify

jimleroyer:
# Summary | Résumé

* ADR to document how secrets are currently used in Notify.
* Template for ADR

      langues officielles?
      fonctionnalité existante?


records/20xx-xx-xx.TEMPLATE.md
```diff
@@ -0,0 +1,29 @@
...
+_TODO: Describe context that triggered the writing of this ADR._
+
+## Options
+
+_TODO: Describe the different options we have._
+
+### Option 1
+
+_TODO: Describe an option._
+
```
patheard, line 22: Maybe have an `## Additional considerations` section here that's optional?  In case there is more info that is useful to understanding the decision.

records/2022-04-14.secrets-usage.md
```diff
@@ -0,0 +1,123 @@
...
+## Methods and libraries
+
+This section enumerates a few methods and libraries to help understand the rest of
+the document to casual readers.
+
+### [itsdangerous](https://itsdangerous.palletsprojects.com/)
+
+This is a library that specializes on signing content in a cryptographic manner.
+A piece of content that is signed with a [HMAC-SHA1 algorithm](https://en.wikipedia.org/wiki/HMAC)
+algorithm. This will not encrypt a message in a one-way manner; we can reverse
```
patheard, line 24:
```suggestion
This is a library that specializes in signing content in a cryptographic manner.
using a [HMAC-SHA1 algorithm](https://en.wikipedia.org/wiki/HMAC). This will not encrypt a message in a one-way manner; we can reverse
```

records/2022-04-14.secrets-usage.md
```diff
@@ -0,0 +1,123 @@
...
+the operation if we have the secret and salt used throughout the signing process.
+Rather, the goal is to guarantee authenticity and integrity of a message.
+
+> In cryptography, an HMAC (sometimes expanded as either keyed-hash message authentication code or hash-based message authentication code) is a specific type of message authentication code (MAC) involving a cryptographic hash function and a secret cryptographic key. As with any MAC, it may be used to simultaneously verify both the data integrity and authenticity of a message.
+
+> HMAC can provide authentication using a shared secret instead of using digital signatures with asymmetric cryptography. It trades off the need for a complex public key infrastructure by delegating the key exchange to the communicating parties, who are responsible for establishing and using a trusted channel to agree on the key prior to communication.
+
+> [Wikipedia](https://en.wikipedia.org/wiki/HMAC)
+
+### [JWT](https://en.wikipedia.org/wiki/JSON_Web_Token)
```
patheard, line 34:
```suggestion
### [JSON Web Token (JWT)](https://en.wikipedia.org/wiki/JSON_Web_Token)
```

records/2022-04-14.secrets-usage.md
```diff
@@ -0,0 +1,136 @@
...
+| verify_codes                 | _code            | No         | No             | The verification codes are generated with 5 random characters first and then bcrypt'ed. <br/>1. [Creating secret code](https://github.com/cds-snc/notification-api/blob/bebd806cad9581101e5cc1d52db2613c1baa1fa9/app/dao/users_dao.py#L23) <br/> 2. [Hashing code with bcrypt](https://github.com/cds-snc/notification-api/blob/9461b26f9582bdea6fe7c68b3d360597e5eab9ba/app/models.py#L1433) |
+
+## Recommandations
+
+1. Produce different salts for each context.
+2. Support within Notify overlapping old and new secrets to safely rotate keys as [suggested by the itsdangerous library](https://itsdangerous.palletsprojects.com/en/2.1.x/concepts/#key-rotation) with no downtime.
+3. Write a script that rotates the crypto-signed fields in the database with latest secrets.
+
```
sastels, line 133: IMO signing things is basically a hassle that's not providing any benefit. I think we should either not sign things or we should encrypt them.

IMO The api key secret should be treated as a password, and hashed. We tell the users "Copy your key to somewhere safe. You will not be able to see it again once you leave this page." which may lead them to believe that we're treating it as a password rather than just storing it in the clear in our database.

records/2022-04-14.secrets-usage.md
```diff
@@ -0,0 +1,136 @@
...
+> In cryptography, an HMAC (sometimes expanded as either keyed-hash message authentication code or hash-based message authentication code) is a specific type of message authentication code (MAC) involving a cryptographic hash function and a secret cryptographic key. As with any MAC, it may be used to simultaneously verify both the data integrity and authenticity of a message.
+
+> HMAC can provide authentication using a shared secret instead of using digital signatures with asymmetric cryptography. It trades off the need for a complex public key infrastructure by delegating the key exchange to the communicating parties, who are responsible for establishing and using a trusted channel to agree on the key prior to communication.
+
+> [Wikipedia](https://en.wikipedia.org/wiki/HMAC)
+
+### [JSON Web Token (JWT)](https://en.wikipedia.org/wiki/JSON_Web_Token)
+
+Notify supports JWT authentication for two scenarios: for authenticating the admin
```
andrewleith, line 36: This is a little unclear for me. Does this mean that admin type users **must** use JWT and cannot the auth header with api key?

records/2022-04-14.secrets-usage.md
jimleroyer: Signing offers integrity and authenticity of the sensitive content across the data pipeline, no? If that is really beneficial and necessary is a good question though. We can ask our security experts to know if that is necessary for a protected B-MM environment. My guess is it might not be required, but every additional security layers add on top of each others à la swiss cheese model. We could deem that data tampering is unlikely within the AWS environment and hence signing is unnecessary (if attackers are at that point, they could do many bad things I assume). I personally like it though, just in case, but we're not married to it.

The API key secret, we don't store in the clear as it is signed, but your point is that it's still decryptable without the secrets. I personally don't have a recommendation for it at the moment because there are pros/cons to be able to track these with the raw values, such as the ability to track these in foreign code repositories. The ideal solution to me would be a one way encryption where we could still recognize API keys in the wild in a reliable manner. I know we started to use prefix for some sensitive values but I wonder if there is not something better.

CC @patheard @mohamed-cds to trigger discussion (should we get rid of signing and what would be the best storage option for API keys with modern security practice?)

mohamed-cds: I also don't see the need to prioritize this enhancement since the added value is low.

general advice is not to rely exclusively on API keys to protect sensitive, critical or high-value resources.

JWT tokens can be used to enforce security controls and API keys to associate requests with users as-well as preventing denial of service since a key is required to interact with the system.

OWASP goes into my depth in the [REST cheatsheet](https://cheatsheetseries.owasp.org/cheatsheets/REST_Security_Cheat_Sheet.html)

jimleroyer: We had a talk offline on this thread. I'll paste here and summarize for proper PR records:

> Pat: I’m thinking about a case where there’s a dispute with someone saying they didn’t set template personalizations with a specific value and we can point back to the signed content to say it wasn’t tampered with
> Steve: that's an interesting point. basically signing gives non-repudiation of the data

> Mohamed: most of the key management controls selected for notify's ATO relate to TLS certificates and doesn't go into detail to the level we're discussing here. So it'll be up to the team to decide. The OWASP guidance usually asks the below questions and whether they're requirements for your app.
> * Confidentiality of data at rest and confidentiality of data in transit.
> * Authenticity of the end device.
> * Authenticity of data origin.
> * Integrity of data in transit.
> * Keys to create the data encryption keys.

> Mohamed: increasing key rotation frequency > non-repudiation since API keys being compromised has a higher probability (in my opinion) than tampering

A few conclusions from this talk:

1. Having signing on content is good on a policy basis, if a legal issue arise. 
2. Rotating secrets on a timely basis is security best practice we should have.
3. Leaking API keys is the one item the most at risk. 
4. Signing API keys offer no real advantage than storing in the clear. This is not content where non-repudiation would be necessary.
5. It would be best to encrypt the API keys in the database, as long as we can identify a pattern out of these to identify leaked/non-encrypted API keys in public and internal code repositories.

records/2022-04-14.secrets-usage.md
jimleroyer: Yes that's the case, but admin user in that case means for HTTP requests moving from the admin to the API. It's not for admin users in the UI. Does that make more sense? And would you like me to reformulate the content in the ADR? Feel free to suggest a better formulation. 🙏

records/20xx-xx-xx.TEMPLATE.md
jimleroyer: I added that!

records/2022-04-14.secrets-usage.md
sastels: re 4. I still think hashing the api keys would be better, as it would mean that there would be no way for anyone to easily recover then from our database / keys (as is the case for the passwords), but encrypting would definitely provide more security than just signing.

re 5. We could consider making the api keys not uuids, but rather random strings with some sort of cds prefix. This would require either allowing old uuid keys to be used as well, or forcing users to re-create their keys.

records/2022-04-14.secrets-usage.md
```diff
@@ -0,0 +1,154 @@
...
+
+```python
+import zlib
+import base64
+
+p = '<paste personalized data here>'
+b64 = p[1:]
+b64_utf = b64.encode('utf-8')
+b64_decoded = base64.urlsafe_b64decode(b64_utf + b'==')
+print(zlib.decompress(b64_decoded))
```
sastels, line 49: We can also get the unsigned content by using `URLSafeTimedSerializer("").loads_unsafe(p)[1]`

records/2022-04-14.secrets-usage.md
jimleroyer: @sastels 

For 4, unfortunately, hashing with itsdangerous means it is easily reversible without the secret and the salt. We could hash it a different way though to avoid that. At that point, we could just encrypt with a prefix or with something that provides a predicable and unique format (if such algo exists -- I don't know!).

For 5, Yes.

mohamed-cds: having a prefix will help tremendously with secret scanning processes since those tools are generally regex based. uuid makes secret scanning difficult

records/2022-04-14.secrets-usage.md
jimleroyer: Ah maybe, it's possible that itsdangerous offers the utility to do it. I just tried it locally though and that returned `(False, None)` on a value where the algo above worked. We can debug tomorrow for curiosity sake. 

I find it nice to be able to decode the value without any help from itsdangerous, to show what's done with no magic. We can include that line for an easier way to do it too.

records/2022-04-14.secrets-usage.md
andrewleith: Ah, understood.  There is a little ambiguity with the term admin for me - it sounded like a user role and not the web app. 

Maybe we could refer to the "Notify admin" as the "Notify admin interface" to be more clear?

> Notify supports JWT authentication for clients accessing the API programmatically as one of two options.  The Notify admin interface uses JWT when communicating with the Notify API.

jimleroyer: Changed!

records/2022-04-14.secrets-usage.md
sastels: The benefit of hashing vs encrypting is that we cannot recover a hashed key (so that its security rests only with the users and not also with our system)

PR #21, opened 2022-05-31
Added results for priority lanes

jzbahrai:
# Summary | Résumé

Added the results after implementing the Priority Lanes ADR.

### DIFF
--- a/records/2022-03-08.priority-lanes.md
+++ b/records/2022-03-08.priority-lanes.md
@@ -73,4 +73,20 @@ We have a low number of priority notifications at current time sent through out 
 
 ## Consequences & Results
 
-_To be completed_.
+Running an hour long performance test where we send 4000 emails/min gives us the following results:
+
+1. Without Priority Lanes
+
+2. With Priority Lanes
+
+The results can be analyzed as follows:
+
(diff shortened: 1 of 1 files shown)


records/2022-03-08.priority-lanes.md
```diff
@@ -73,4 +73,23 @@ We have a low number of priority notifications at current time sent through out
...
+
+
+The results can be analyzed as follows:
+
+With Priority Lanes switched on we see the time for a 'priority' message reduces from 18 minutes down to 10 seconds.
+
+The other headings are as follows:
+
+* Redis - Time spent after the POST is completed but before it is saved to our DB. 
+* Delivery - Time taken by AWS to send the message, upto when we get a callback informing us the message is delivered. It is likely that our callback is extremely unoptimized here, causing a skew in the results.
```
jimleroyer, line 92:
```suggestion
* Delivery - Time taken by AWS to send the message, up to when we get a callback informing us the message is delivered. We don't have control over the delivery callback and the timing of that response does not depend on us. It usually takes much longer to receive that callback than the actual email reception so we cannot rely too much on that metric.
```

jimleroyer (commented):
Thanks for adding the results to the ADR. I left a suggestion about a typo and rephrase our expectations around the delivery response we get from AWS.

PR #22, opened 2022-06-01
Baseline Notify

sastels:
# Summary | Résumé

ADR to determine how to do baseline performance testing


sastels:
I don't think we need this anymore... closing for now but will keep the branch around...

PR #29, opened 2022-12-13
Upgrading the database

sastels:
# Summary | Résumé

https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/1022

### DIFF
--- a/records/2022-12-13.upgrade-database.md
+++ b/records/2022-12-13.upgrade-database.md
@@ -1,0 +1,60 @@
+# TITLE
+
+Date: 2022-12-13
+
+## Status
+
+**DRAFT**.
+
+## Context
+
+[Staging](https://ca-central-1.console.aws.amazon.com/rds/home?region=ca-central-1#database:id=notification-canada-ca-staging-cluster;is-cluster=true;tab=configuration) and 
+[production](https://ca-central-1.console.aws.amazon.com/rds/home?region=ca-central-1#database:id=notification-canada-ca-production-cluster;is-cluster=true;tab=configuration) are both running aurora-postgresql 11.13 (from 2021-08-12). The latest version of PostgeSQL available on AWS is 14.5. The latest version of PostgeSQL is 15.1.
+
+AWS recommended that we upgrade to at least 11.17 by July 23, 2022.
+
+## Options
+
+According to [Upgrading the PostgreSQL DB engine for Amazon RDS](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html) we cannot go straight to 14.5. Our options are:
+- go to 11.17
+- go to 12, ie 12.8, 12.9, 12.10, 12.11, or 12.12
+- go to 13.14 (AWS maximum recommended upgrade)
+
+According to [AWS](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html):
+
+**Major version upgrades**
+
+Major version upgrades can contain database changes that are not backward-compatible with existing applications. As a result, you must manually perform major version upgrades of your DB instances. You can initiate a major version upgrade by modifying your DB instance. However, before you perform a major version upgrade, we recommend that you follow the steps described in [Choosing a major version upgrade for PostgreSQL](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html#USER_UpgradeDBInstance.PostgreSQL.MajorVersion). During a major version upgrade, Amazon RDS also upgrades all of your in-Region read replicas along with the primary DB instance.
+
+**Minor version upgrades**
+
+In contrast, minor version upgrades include only changes that are backward-compatible with existing applications. You can initiate a minor version upgrade manually by modifying your DB instance. Or you can enable the Auto minor version upgrade option when creating or modifying a DB instance. Doing so means that your DB instance is automatically upgraded after Amazon RDS tests and approves the new version. If your PostgreSQL DB instance is using read replicas, you must first upgrade all of the read replicas before upgrading the primary instance. If your DB instance is in a Multi-AZ deployment, then the writer and any standby replicas are upgraded simultaneously. Therefore, your DB instance might not be available until the upgrade is complete. For more details, see [Automatic minor version upgrades for PostgreSQL](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html#USER_UpgradeDBInstance.PostgreSQL.Minor). For information about manually performing a minor version upgrade, see [Manually upgrading the engine version](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.Upgrading.html#USER_UpgradeDBInstance.Upgrading.Manual).
+
+### 11.17
+
+Likely not risky.
+
+### 12
+
+There are [breaking changes](https://www.postgresql.org/docs/12/release-12.html#id-1.11.6.18.4) from 11 to 12. Unknown if any affect us.
+
+### 13.14
+
+There are also [breaking changes](https://www.postgresql.org/docs/13/release-13.html#id-1.11.6.14.4) from 12 to 13. Unknown if any affect us.
+
+
+## Additional considerations
+
+It appears that even a minor upgrade requires database downtime. We should have a proper maintenance window where Notify admin displays a "Notify is unavailable" page and the api returns a "Notify is unavailable" error message. 
+
+Note that our [terraform](https://github.com/cds-snc/notification-terraform/blob/main/aws/rds/rds.tf#L72) sets the engine version to 11.9, but adds the engine version to the `ignore_changes` section. So we do not have to change our terraform to change the engine version.
+
+However we have a custom [RDS cluster parameter group](https://github.com/cds-snc/notification-terraform/blob/main/aws/rds/rds.tf#L41) that is using `aurora-postgresql11` for the `family`. If we move to 12 or 13 we probably have to change the family here.
+
+## Decision
+
+_TODO: Describe the way forward._
+
+## Consequences
+
+_TODO: Describe foreseen and felt consequences of the decision (possible after 1-3 months)._


records/2022-12-13.upgrade-database.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Option D: Replication-based approach using pglogical
+This approach use the `pglogical` database extension along with database replication sets (A replication set is way to mark a collection of tables to be replicated.). In many ways this is very similar to Option C, however in this case once the new database is populated it stays up to date faster, and there is no "wait for last entries to transfer" downtime. Rather we just have to turn off Notify, switch to the new database, and turn Notify back on (and then turn off the extension). There are restriction on what database configurations are acceptible, however, and this is a more manual process than using DMS.
+
+### Discussion
+
+We want to minimize downtime and maximize probability of success. 
+- There probably isn't too much to be gained performance-wise by just moving to 11.16, so we should move to 12.12 or 13.14. Just moving to 12.12 will increase the probability of success, at the expense of having to do anothe migration at a future point.
```
whabanks, line 65: "at the expense of having to do anothe migration at a future point" - Missing 'r' in another

records/2022-12-13.upgrade-database.md
```diff
@@ -0,0 +1,104 @@
...
+- Note that one benefit to just a minor upgrade would be that we could be confident that any issues that arise are due to our migration and not to the version change. So we could use a 11.16 upgrade as an opportunity to gain familiarityt with the migration process.
```
whabanks, line 66: "opportunity to gain **familiarityt**" - Extra 't' on the end.

whabanks (changes requested):

records/2022-12-13.upgrade-database.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Upgrade approaches
+
+From [Upgrade your Amazon RDS for PostgreSQL or Amazon Aurora PostgreSQL database, Part 1: Comparing upgrade approaches](https://aws.amazon.com/blogs/database/part-1-upgrade-your-amazon-rds-for-postgresql-database-comparing-upgrade-approaches/) there are a few approaches we can use.
+
+### Option A: In-place upgrade
+This is in some ways the easiest, in that we essentially tell AWS to upgrade the database and it does it. The downside is that the database will be unavailable for most of the upgrade time, and hence Notify will be down. Also, if something goes wrong we would have to restore the database from a backup (and hope that works).
+
+### Option B: Snapshot approach
+In this approach we would make a copy of the database and upgrade that copy, then switch the app to use the new database. The benefit is that if something doesn't work we can quickly switch back to the old database. This approach still requires downtime during the process though, since data added to the old database would not be in the new database.
```
jzbahrai, line 53: we can in theory use the redis url approach and not have the celery workers read anything - this way we just switch on the celery workers after the migration and they then write to the new DB.

records/2022-12-13.upgrade-database.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Discussion
+
+We want to minimize downtime and maximize probability of success. 
+- There probably isn't too much to be gained performance-wise by just moving to 11.16, so we should move to 12.12 or 13.14. Just moving to 12.12 will increase the probability of success, at the expense of having to do another migration at a future point.
+- Note that one benefit to just a minor upgrade would be that we could be confident that any issues that arise are due to our migration and not to the version change. So we could use a 11.16 upgrade as an opportunity to gain familiarity with the migration process.
+- Option A is probably the worst on both scores. 
+- Option B give us the ability to quickly switch to the old database if things go wrong, but the downtime will be the same (likely a bit longer because we need to also take the snapshot). 
+- Options C and D probably have the same downtime for us (if we do this on a Saturday evening, for example, there will not be may notifications sent during Option C's "wait for last entries to transfer" window)
+- Option C may be less manual and more documented / supported. Both approaches require multiple manual steps however.
```
jzbahrai, line 70: option c seems the most straightfoward

patheard: I'd vote for option C as well since it seems like AWS would take care of most of the complexity for us.

records/2022-12-13.upgrade-database.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Discussion
+
+We want to minimize downtime and maximize probability of success. 
+- There probably isn't too much to be gained performance-wise by just moving to 11.16, so we should move to 12.12 or 13.14. Just moving to 12.12 will increase the probability of success, at the expense of having to do another migration at a future point.
+- Note that one benefit to just a minor upgrade would be that we could be confident that any issues that arise are due to our migration and not to the version change. So we could use a 11.16 upgrade as an opportunity to gain familiarity with the migration process.
```
patheard, line 66: I like this idea.  What if we went to the highest `11.x` minor which is `11.18` as the test?  

Then if it works, we could work through the upgrades sequentially going through each major one at a time:
```mermaid
  graph LR;
      11.13-->11.18;
      11.18-->12.13;
      12.13-->13.9;
      13.9-->14.5;
```

records/2022-12-13.upgrade-database.md
```diff
@@ -0,0 +1,104 @@
...
+
+### 11.17
+
+Likely not risky.
+
+### 12
+
+There are [breaking changes](https://www.postgresql.org/docs/12/release-12.html#id-1.11.6.18.4) from 11 to 12. Unknown if any affect us.
```
jimleroyer, line 39: Can we use this ADR to investigate the breaking changes and see if that could affect us? 

We could run a local database the target versions and see if that is problematic. Which on that note, I always run the latest locally... so I know some things work fine with GCNotify.

records/2022-12-13.upgrade-database.md
```diff
@@ -0,0 +1,104 @@
...
+
+### 13.14
+
+There are also [breaking changes](https://www.postgresql.org/docs/13/release-13.html#id-1.11.6.14.4) from 12 to 13. Unknown if any affect us.
```
jimleroyer, line 43: Let's use this ADR to enumerate the ones that might be risky.

records/2022-12-13.upgrade-database.md
jimleroyer: We might be able to leverage RDS Proxy for such approach:

* https://aws.amazon.com/blogs/database/improving-application-availability-with-amazon-rds-proxy/
* https://catalog.us-east-1.prod.workshops.aws/workshops/098605dc-8eee-4e84-85e9-c5c6c9e43de2/en-US/lab8-fault-tolerance/proxy

> Failover occurs when the primary database instance becomes inaccessible and another instance takes over as the new primary. This disrupts client connections. Failovers can be planned, when they are induced by administrative actions such as a rolling upgrade, or unplanned, when they occur due to failures. In both cases, you want to reduce downtime to minimize client disruption.

> Amazon RDS has multiple high availability options, and RDS Proxy provides failover recovery benefits for each option. The Amazon RDS Multi-AZ option is a primary and standby configuration with synchronous replication. Aurora offers two high availability options—up to 15 asynchronous replicas and multi-master. All Aurora replicas are available for read-only access. You can elect and promote any replica as a new primary for the cluster in the event of failover without any data loss. Aurora multi-master offers multiple active writers.

> With all of these options, when a failover occurs, the client has to detect the connection failure, discover a new primary, and reconnect to it as quickly as possible. With RDS Proxy, the applications can avoid the complexity associated with failovers and experience faster recovery (in as few as 3 seconds). With Aurora, where failovers are faster, DNS propagation delay is the largest contributor to your overall failover time. RDS Proxy actively monitors database instances and automatically connects clients to the right target. It also maintains idle client connections through database failover without dropping them. Idle connections in this context are connections that don’t have outstanding requests.

records/2022-12-13.upgrade-database.md
jimleroyer: Good testing approach indeed.

records/2022-12-13.upgrade-database.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Upgrade approaches
```
jimleroyer, line 45: Can you please estimate an effort scale to each of these options, with 1 being the easiest and 5 the hardest? That would provided a clearer picture between option C and D for example.

records/2022-12-13.upgrade-database.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Discussion
+
+We want to minimize downtime and maximize probability of success. 
+- There probably isn't too much to be gained performance-wise by just moving to 11.16, so we should move to 12.12 or 13.14. Just moving to 12.12 will increase the probability of success, at the expense of having to do another migration at a future point.
+- Note that one benefit to just a minor upgrade would be that we could be confident that any issues that arise are due to our migration and not to the version change. So we could use a 11.16 upgrade as an opportunity to gain familiarity with the migration process.
+- Option A is probably the worst on both scores. 
+- Option B give us the ability to quickly switch to the old database if things go wrong, but the downtime will be the same (likely a bit longer because we need to also take the snapshot).
```
jimleroyer, line 68: We should probably take a snapshot in all options, right? (but I understand with option B and the lose no data approach, it would need a maintenance period).

records/2022-12-13.upgrade-database.md
jimleroyer: It seems the best. What is the required effort with this option?

patheard: Honestly not sure as I haven't done an upgrade of this scale before.  We'll have to do more research and test it out with our Staging upgrade.

jimleroyer: A gross estimate could work, relative to other options (not time related).

records/2022-12-13.upgrade-database.md
sastels: maybe... this is different from the redis migration where in redis data comes in and then leaves, and it's easy to copy over anything that seems stuck. Here data just circles around inside the database, essentially - there's no way to "drain" it as did for redis 🤔 .

records/2022-12-13.upgrade-database.md
jimleroyer: Good reference @sastels , it gives a better perspective. We can include it and compare to other commands listing?

idavidmcdonald:
Howdy from the UK GOV.UK Notify! Here to share what we are up to in a similar area.

So the GOV.UK PaaS where we are hosted is being retired. Currently our database is an RDS postgres 11 database running in the GOV.UK PaaS AWS account. We will need to end up running our own RDS postgres in our own AWS account. We've ended up choosing to use AWS DMS to migrate our data out of the GOV.UK PaaS AWS account into our new RDS database in our account.

We've done a bunch of dry runs and learning. We are aiming for minimal downtime if possible and think DMS is worth it. Our main alternative is taking a database snapshot, sharing it across accounts (potentially with some KMS encryption challenges to be solved) and then restoring a new database from the snapshot. This option would likely mean downtime for our users of over an hour (could be more).

Our main takeaway is that DMS is not actually that simple and comes with a far bit of complexity. A few particulars

1. We don't trust DMS to copy the database schema across to our new database. We looked at trusting it do to this and there were two issues. Firstly, it doesn't copy all of the schema across so you are then left unpicking what bits of the schema it copied across and what it didn't and then applying manual schema changes on top. Secondly, it generally gets the schema conversion right but we found a couple of things (I don't recall them off the top of my head, but potentially to do with default values) didn't get copied across. 

2. There is a bunch of work to set it up. Not tons, but not insignificant. A fair bit of reading through the AWS docs for something that people here aren't very familiar with. People are much more familiar with snapshots or the upgrade process that AWS offers.

3. We've had to put more validation in place to make sure that all the records have been transferred correctly and we've entered a point of replication where the source and target database are in sync enough to stop traffic going to our source database, let replication finish and then point our apps at our new database.

The headline is that DMS seems OK for moving our database from one AWS account to another, but we don't think it is worth the risk/complexity involved for upgrading your postgres versions. GOV.UK Pay over here looked into this themselves and also spoke to us about it as they have several databases they need to upgrade the version of. Based on their own findings, and chatting to us, they decided not to use DMS. 


However... if we are going to be using DMS anyway to move data out of our postgres 11 database and into a new postgres database, why not make that new database postgres 15! That is our plan and we've done the following to give us confidence it will work
- we've practiced this using DMS in our preview environment with our target database being postgres 15
- we've run our unit tests for the API locally against postgres version 15 (the amount of changes needed were minimal https://github.com/alphagov/notifications-api/pull/3800)
- we've used DMS in our preview environment to transition to postgres 15 and we are going to leave it running on postgres 15 for a while to see if there are any further issues we can catch

I see you are considering `pglogical`. We didn't really do any looking into it, but that we weren't so aware of it, rather than we looked into it and thought it is a bad option.

We've done a couple of postgres version upgrades in the past. My view is that its best to go a few versions in one session. Given the amount of work involved, in particular communicating with users to make them aware of downtime. Its better to once every few years have a bit of large downtime than have to go through this hoop every year. But that is quite a personal opinion.

Best of luck and let us know if you have any questions or bits to clarify!

records/2023-12-19.upgrade-database-major-version.md
```diff
@@ -0,0 +1,160 @@
...
+
+1. The notify databases does not have logical replication enabled. This is easily fixed by adding the enable flag in Terraform, however it requires a database restart to take effect, thus creating an additional 10 seconds of downtime.
+
+2. The blue/green deployment strategy does not support the usage of an RDS proxy. Notify uses an RDS proxy to improve database performance, since the RDS proxy is better able to manage a large number of concurrent connections to the database from the celery workers and API lambdas. It will be necessary to remove the RDS Proxy temporarily so that the blue/green deployment can be performed.
+
+## Steps
+
+The following upgrade process was established to mitigate the aforementioned problems while performing the database upgrade.
+
+[15.5 Database Migration Scripts](https://github.com/cds-snc/notification-attic/tree/main/scripts/database-migration)
```
jimleroyer, line 75:
```suggestion
Steps that could be automated are supported by our [PostgreSQL 15.5 Migration Scripts](https://github.com/cds-snc/notification-attic/tree/main/scripts/database-migration). Please refer to its README for instructions and how it works.
```

jimleroyer (approved):
LGTM!

whabanks (approved):

sastels (commented):
New changes LGTM too! I shall merge this in 🎉

PR #30, opened 2023-01-24
Malicious Content Scanning

smcmurtry:
# Summary | Résumé

An ADR to help us move forward with the work to scan attachments for malicious content.


records/2023-01-24.malicious-content-scanning.md
```diff
@@ -0,0 +1,61 @@
...
+
+We have explored using the `asyncio` library in python to create a new process to call the scan-files api [here](https://github.com/cds-snc/notification-document-download-api/pull/97/files#diff-a842b15bd72bea41437e64bda92d12262c8eca35af7e8adf111bd520fe7c3010R1). Then once the result is returned, we could write that result to the S3 object as a tag which can be read later. However, this approach did not work to reduce the response time for the Document Download API.
+
+It seems that it is [not possible to create background processes inside flask](https://flask.palletsprojects.com/en/2.2.x/async-await/#background-tasks).
+
+## Options
+
+Here are some of the options for scanning email attachments with the scan-files project.
+
+### Migrate Document Download API from Flask to Quart
```
jimleroyer, line 29: An option to migrate from Flask to Flask async could still be considered, as a complete migration to Quart might be more effort and Flask async might be enough to work for us.

records/2023-01-24.malicious-content-scanning.md
```diff
@@ -0,0 +1,61 @@
...
+
+The SRE team has created the [scan-files](https://github.com/cds-snc/scan-files) project, which is a tool to help platform products scan files for malicious content. It includes terraform for [S3 scan object](https://github.com/cds-snc/scan-files/tree/main/module/s3-scan-object), which is sets up a lambda function that watches a given S3 bucket and scans each new file. The scan verdict is then written as a tag to the S3 object, and can be read when the S3 object is accessed.
+
+We have [tried getting the S3 scan object lambda function to run on our S3 bucket](https://github.com/cds-snc/notification-terraform/pull/600), but that did not work because we are using a different encryption key for each file in S3. This is because our S3 bucket for file uploads is public, and we need to prevent a user from being able to download and decrypt files they should not have access to.
```
jimleroyer, line 17: We checked during the meeting if our S3 bucket was public and our assumption was wrong: it's private. We actually deliver link from dd-api as a proxy. Someone who would want to download a file he shouldn't have access would have to guess the S3 file UUID and the encryption key.

records/2023-01-24.malicious-content-scanning.md
```diff
@@ -0,0 +1,104 @@
...
+Eventually the email sending job is processed by Celery, and the following occurs: 
+```mermaid
+flowchart TD
+    A[Celery calls `send_email_to_provider` fn]
+    B{Does email contain a file attachment?}
+    C1{request to dd-api to \n/check-scan-verdict endpoint}
+    C2[Send email as usual]
+    D[Wait for 5 minutes and try again]
+    E[Send email]
+    F[Do not send email\n Set notification status to 'malware-scan-failure']
```
patheard, line 60: Out of curiosity, how is this `malware-scan-failure` communicated back to the user who tried to send the email with the malicious attachment?

Also, it's possible that a few different [scan verdicts](https://github.com/cds-snc/scan-files/blob/233cca8d3f400d35616d76b71479ddf0238d7554/api/models/Scan.py#L16-L23) could be returned:
```
IN_PROGRESS = "in_progress"
CLEAN = "clean"
MALICIOUS = "malicious"
ERROR = "error"
UNABLE_TO_SCAN = "unable_to_scan"
```

records/2023-01-24.malicious-content-scanning.md
```diff
@@ -0,0 +1,104 @@
...
+    D-->E
+```
+Eventually the email sending job is processed by Celery, and the following occurs: 
+```mermaid
+flowchart TD
+    A[Celery calls `send_email_to_provider` fn]
+    B{Does email contain a file attachment?}
+    C1{request to dd-api to \n/check-scan-verdict endpoint}
+    C2[Send email as usual]
+    D[Wait for 5 minutes and try again]
```
patheard, line 58: I realize that I'm the one that steered you down the path towards async scanning, but that may have been a mistake if you're going to delay the email send until a verdict is returned.  Apologies for not realizing this sooner, but if you used a synchronous invoke of the Scan Files API from the Notify code, it would accomplish the same thing.

That being said, the risk would be if there's a problem with the Scan Files API that prevents it from scanning, so some timeout logic would be needed on the API call.

records/2023-01-24.malicious-content-scanning.md
smcmurtry: For the statuses "malicious" and "suspicious":

My thinking is that the email would be marked as "Failed" and there would be an explanation like "Malware scan failure" - we don't have the exact content yet. That message would appear as the red text in the image below:
Screen Shot 2023-02-07 at 11 08 55 AM

For "in_progress": 
Try again in 5 minutes.

For all other statuses:
I think we would just send the email as usual so that we don't block sending if scan-files goes down or there is a weird error. But maybe we could have an alert to tell us if we are getting a lot of errors.

patheard: Ah that makes sense 👍🏻

records/2023-01-24.malicious-content-scanning.md
smcmurtry: That's okay! I think the S3 bucket solution is still better for us. If we called the scan-files api directly and waited for a response, that wait would have to happen in Celery, so that we don't take minutes to return an API response to the user. And then in Celery, the waiting for a response could have worse performance that making a request to "/check-scan-verdict" on the dd-api, quickly getting the response, and then trying again later if the scan is still in progress.

Though I am now thinking that retrying after 5 minutes may be too long. If the scan is often less than 1 minute, we could try again at 1 minute, 2 minutes, etc

patheard: Gotcha and yeah, most responses are returned within a second or two.  It's only in the case of a cold-start on our scanning API that we see the response take ~30 seconds.

jimleroyer: @smcmurtry 💯 We like the async nature of it so we don't block on process or pollute it.

records/2023-01-24.malicious-content-scanning.md
jimleroyer: >I think we would just send the email as usual so that we don't block sending if scan-files goes down or there is a weird error. But maybe we could have an alert to tell us if we are getting a lot of errors.

+1 for raising an alarm if we get many errors.

Can we please add more paths to the decision tree diagram in case that malware scanning returns an error, or is down? to demonstrate case when we actually send the emails even in the case of uncertainty.

smcmurtry: I added the error case and also the case where the scan is taking too long because `scan-files` is down.

records/2023-01-24.malicious-content-scanning.md
```diff
@@ -0,0 +1,110 @@
...
+
+However, we would need to do a significant amount of testing to determine the performance and reliability of this solution.
+
+### Implement a new queue in Document Download API to start background tasks
+
+Use Celery to store a queue of tasks to query the scan-files API and add a tag with the scan verdict to the S3 object. One downside of this is that it would introduce more complexity to the document download api, since Celery is not set up there yet.
+
+### Create a new api/lambda function to perform the task 
+
+Another possibility? T
```
jimleroyer, line 98: This section seems incomplete.

records/2023-01-24.malicious-content-scanning.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Migrate Document Download API from Flask to Quart
+
+Flask does not have good async support, and [they suggest using Quart for better async support](https://flask.palletsprojects.com/en/2.2.x/async-await/#when-to-use-quart-instead). 
+
+It does seem like Quart [adds support for background tasks](https://quart.palletsprojects.com/en/latest/how_to_guides/background_tasks.html) and would solve our problem.
+
+It seems like migrating from Flask to [Quart](https://github.com/pallets/quart) is not very difficult (see [quart docs](https://quart.palletsprojects.com/en/latest/how_to_guides/flask_migration.html)).
+
+Update: We tried using Quart and it introduced some compatibility problems with [notification-utils](https://github.com/cds-snc/notification-utils).
```
jimleroyer, line 83:
```suggestion
Unfortunately, as we tested Quart, it introduced some compatibility problems with [notification-utils](https://github.com/cds-snc/notification-utils). It makes this option less attractive for the lift it requires, on top of migrating.
```

records/2023-01-24.malicious-content-scanning.md
```diff
@@ -0,0 +1,110 @@
...
+
+Use Celery to store a queue of tasks to query the scan-files API and add a tag with the scan verdict to the S3 object. One downside of this is that it would introduce more complexity to the document download api, since Celery is not set up there yet.
+
+### Create a new api/lambda function to perform the task 
+
+Another possibility? T
+
+## Additional considerations
+
+_TODO: Describe extras in here._
```
jimleroyer, line 102: We can remove this section if we won't use it.

records/2023-01-24.malicious-content-scanning.md
```diff
@@ -0,0 +1,110 @@
+# Malicious Content Scanning
+
+Date: 2023-01-24
+
+## Status
+
+**DRAFT**.
+
+_VALUES: (DRAFT, IN REVIEW, APPROVED, REJECTED)_
```
jimleroyer, line 9: Pretty much Approved by now. 😄

records/2023-01-24.malicious-content-scanning.md
smcmurtry: I'll remove

jimleroyer (approved):
LGTM!

PR #32, opened 2023-01-27
Bounce rate automated emergency brake

andrewleith:
# Summary | Résumé

ADR to lay out how to implement an automated emergency brake for bounce rate.


records/2023-01-26.bounce-rate-automated-emergency-brake.md
```diff
@@ -0,0 +1,83 @@
...
+### Calculation
+AWS calculates the bounce rate as the number of hard bounces per the total number of sent emails, expressed as a percentage:
+![bouncerateformula](./diagrams/2023-01-26.bounce-rate-automated-emergency-brake/bouncerate.png)
+
+### AWS Thresholds
+- A bounce rate is considered **Healthy** when it remains below 5%.
+- A bounce rate is considered **Under review** when it is between 5% and 10% (>=5% and <10%).
+- A bounce rate is considered **At risk** when it is greater than or equal to 10%.
+
+### Why the bounce rate is important
```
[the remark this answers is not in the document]
andrewleith: Done! 👍

records/2023-01-26.bounce-rate-automated-emergency-brake.md
```diff
@@ -0,0 +1,83 @@
...
+### Notify a service when it is within [bounce-rate-warning-percentage] of the [bounce-rate-threshold]
+When a service is approaching the [bounce-rate-threshold], notify the service owner so they can take action to reduce their bounce rate.  The message should contain a link to the service's dashboard where they can see their bounce rate and take action to reduce it.
+
+#### Minimum volume threshold
+Low-volume senders won't have a large impact on the bounce rate, and notifying them directly may lead to many support calls. In order to limit this, only notify services that are sending more than [bounce-rate-service-volume-minimum] emails in the [bounce-rate-window].
+
+### Suspend a service when it reaches the [bounce-rate-threshold]
+When a service reaches the [bounce-rate-threshold], suspend their service to prevent them from sending any more emails. Notify the service owner that their service has been suspended and provide them with a link to the service's dashboard where they can see their bounce rate and take action to reduce it.
+
+:question: _How to decide when it is appropriate to unsuspend a service?_
```
jimleroyer, line 50: My current take is that this will require a process of a) communicating with users through support to tell them what the issue is and provide a fix on their end first, b) looking at our AWS account bounce rate and determine if we are in a danger zone, c) once users fixed their distribution list, review their fix and wait for Notify bounce rate to go down if in a dangerous zone to reactivate the service, d) monitor the service and their bounce rate in the following days to check that their fix was effective. Would that make sense?

records/2023-01-26.bounce-rate-automated-emergency-brake.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Add bounce rate information to the service dashboad
+Add the bounce rate to the service dashboard so that service owners can see their bounce rate and proactively read the guidance on it so they can prevent their bounce rate from rising in the first place.
+
+### Provide bounce rate guidance
+Provide guidance to service owners to explain bounce rate concepts and how they apply to their service.  Provide guidance on how services can keep their bounce rate low. 
+
+### bounce-rate-threshold calculation options
+When calculating a service's bounce rate, we can either consider them in isolation or in aggregate with other services.  The bounce rate is calculated as a percentage of the total number of emails sent.  If we consider a service in isolation, then the bounce rate will be calculated as a percentage of the total number of emails sent by that service.  If we consider a service in aggregate, then the bounce rate will be calculated as a percentage of the total number of emails sent by all services.
```
jimleroyer, line 59: We do want a bounce rate isolated to a service, this is our need. But we could actually calculate service isolated and global bounce rate, the latter would be useful for admins.

andrewleith: Yeah I was trying to balance understandability with usefulness here and also thinking about bounce rates in tow categories: 1) those that affect our overall rate and 2) those that that don't.  

I think the best bet will be to base the dashboard display on the bounce rate in relation to the service's total send rate as this will make the most sense to a user.  In terms of notifying/suspending services however, we could do that for services with high individual bounce rates and ALSO are actually impacting the overall bounce rate.  This would reduce support requests on our end with the trade-off of letting lower-volume services keep high bounce rates while protecting the overall Notify platform by focusing on the most problematic services.

I think the key to this will be devising the `bounce-rate-service-volume-minimum` and `bounce-rate-window` values to ensure we focus on the right services.  I will run a some calculations and add some tables to the ADR to try to illustrate this.

records/2023-01-26.bounce-rate-automated-emergency-brake.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Staggered roll-out
+To prevent suspending services that may not actually be causing issues with the overall bounce rate, it is suggested to roll out this feature in stages.  The first stage would include tracking bounces, calculating bounce rates, and displaying them on service dashboards. It would also include bounce rate guidance for services. Once we are sure our calculations work satisfactorily with live data, we would move to the second stage which will include notifying services when they are approaching the bounce rate threshold and suspending services when they reach the threshold.
+
+## Additional considerations
+
+### How GC Notify handles bounce notifications from SES
+GC Notify uses a lambda function named `ses-to-sqs-email-callbacks` to handle bounce notifications sent back from SES.  When a callback comes in, the lambda function parses the bounce notification and sends it a queue to be picked up and run by the celery task `process-ses-result`.  The `process-ses-result` task updates the status of the notification in the database including whether it was a hard or a soft bounce.
```
jimleroyer, line 73: This explains well how the signal for bounce notification might come into the Celery task. We could also explain a) how we plan to record it here, if we should leverage Redis or other tool, b) what's the time period we should use, c) how do we map the service to the received bounce message.

records/2023-01-26.bounce-rate-automated-emergency-brake.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Provide bounce rate guidance
+Provide guidance to service owners to explain bounce rate concepts and how they apply to their service.  Provide guidance on how services can keep their bounce rate low. 
+
+### bounce-rate-threshold calculation options
+When calculating a service's bounce rate, we can either consider them in isolation or in aggregate with other services.  The bounce rate is calculated as a percentage of the total number of emails sent.  If we consider a service in isolation, then the bounce rate will be calculated as a percentage of the total number of emails sent by that service.  If we consider a service in aggregate, then the bounce rate will be calculated as a percentage of the total number of emails sent by all services.
+
+#### Option 1: Consider a service in isolation
+Considering a service in isolation would mean calculating the bounce rate solely based on an individual services activity.  The benefit of this is that the calculation would be straight-forward and easy to understand by the service.  The downside is that even if a particular service has a high bounce rate, it may not have a large impact on the overall bounce rate.
```
jimleroyer, line 62: Overall, if all services are below their 10% bounce rate, we can confidently say that the global bounce rate calculated on the total email should be below 10% as well, right?

records/2023-01-26.bounce-rate-automated-emergency-brake.md
```diff
@@ -0,0 +1,83 @@
...
+
+#### Option 1: Consider a service in isolation
+Considering a service in isolation would mean calculating the bounce rate solely based on an individual services activity.  The benefit of this is that the calculation would be straight-forward and easy to understand by the service.  The downside is that even if a particular service has a high bounce rate, it may not have a large impact on the overall bounce rate. 
+
+#### Option 2: Consider a service in aggregate
+Considering a service in aggregate would mean calculating the bounce rate based on the total number of emails sent by all services.  The benefit of this is that the calculation would be more accurate and would reflect the impact of a particular service on the overall bounce rate.  The downside here is that the calculation would be more complex and may be difficult to understand by the service, since for example even if they had an extremely high bounce rate, we might not indicate it as problematic if it has little impact on the overall bounce rate.
```
jimleroyer, line 65: OK I see what you are going for here, trying to allow for a higher bounce rate for an individual services if the total volume of notification can amortize it. 

You are right in saying this will be a difficult rule to explain to our users. We'd have to explain this in our content and I am not sure it is best to expose influencing metrics transparently to our users (e.g. global bounce rate).

This feels overly complex and potentially, could lead to use cases where a service that has a bounce rate of only 5% to get suspended because the global bounce rate was previously boosted up by another service that was deemed OK to reach a 20% bounce rate with amortization. I think there might be ways to enable amortization, but not in a dynamic/unplanned manner, which could affect the services ecosystem.

jimleroyer (commented):
This is well written and I quite like the provided context with all different options/discussions we had so far. 👍

records/2023-01-26.bounce-rate-automated-emergency-brake.md
jimleroyer: Oh I just read that reply. This might be related to another comment I just left and I think it could be problematic and complex to implement. We can discuss in person to have a higher communication bandwidth.

andrewleith: Yes let's discuss.  I ran some numbers comparing the idea of individual vs overall bounce rates and you are right it is overly complex for no real added benefit.  I'm going to re-work the ADR to remove this more complex idea.

records/2023-01-26.bounce-rate-automated-emergency-brake.md
andrewleith: Agreed, after considering this some more, this won't provide much additional benefit for us in the end, and its best to keep our each user's individual bounce rate low no matter what any other user is doing on the platform.  I'll re-work this and simplify it.

records/2023-01-26.bounce-rate-automated-emergency-brake.md
andrewleith: Added extra details to ADR.

records/2023-01-26.bounce-rate-automated-emergency-brake.md
```diff
@@ -31,47 +33,99 @@ AWS calculates the bounce rate as the number of hard bounces per the total numbe
...
+
+#### Calculating the bounce rate
+Since the bounce rate can nearly be calculated as-is from the data currently pulled live into the dashboard, it is possible to calculate the bounce rate on the fly in the dashboard with minimal code changes.  This avoids the need to explicitly store the bounce rate in the database or redis.
+
+#### Changes to the dashboard
+The dashboard already pulls weekly and daily sending statistics including sending totals and total failures.  This could be extended to pull data also by `[bounce-rate-window]` to enable bounce rate calculation.  Failure data will need to be broken down further than it currently is to include the bounce type.  
+
+#### Monitoring bounce rates
+Bounce rates can be monitored by a periodic celery task that polls the database for bounce rates, and notifies services that are over `[bounce-rate-warning-percentage]` and suspends services that are over `[bounce-rate-critical-percentage]`.  This task should be run several times a day to ensure that services are notified and suspended as soon as possible after they reach one of these limits.  A query to get service bounces rates could look something like this:
```
jimleroyer, line 84: Makes sense to check bounce rates with a repeating Celery task! 

What about checking each services bounce rate from Redis instead of the database? This would reduce the query load on the database, as we'd need to run this aggregate query often (the more the better, i.e. every minute).

records/2023-01-26.bounce-rate-automated-emergency-brake.md
```diff
@@ -31,47 +33,99 @@ AWS calculates the bounce rate as the number of hard bounces per the total numbe
...
+- :warning: There does not appear to be code to record synchronous bounces in this method.
+
+## Proposed values
+
+The values below can be used as a starting point for this feature.  They should be reviewed and adjusted as needed as the feature is tested and rolled out.
+
+| Parameter                                       | Value    | Notes                                                                                                                        |
+|-------------------------------------------------|----------|------------------------------------------------------------------------------------------------------------------------------|
+| bounce-rate-warning-percentage                  |       5% | AWS indicates a 5% bounce rate as under review, so we should warn no higher than this                                        |
+| bounce-rate-critical-percentage                 |       9% | AWS indicates a 10% or higher bounce rate as at risk, so we should suspend before this                                       |
```
jimleroyer, line 125: We can have this at 10% I think.

records/2023-01-26.bounce-rate-automated-emergency-brake.md
```diff
@@ -31,47 +33,99 @@ AWS calculates the bounce rate as the number of hard bounces per the total numbe
...
+| bounce-rate-window                              | 24 hours | This should be at a minimum 24 hours, but could be a longer period                                                           |
+| bounce-rate-warning-notification-volume-minimum |     2500 | This should be a significantly high number to ensure we only notify users with enough volume to sway the overall bounce rate |
```
jimleroyer, line 127: I'd shoot with 1,000 but 2,500 could be fine too.

records/2023-01-26.bounce-rate-automated-emergency-brake.md
andrewleith: We certainly could.  I wasn't really worrying about the DB load since the admin dashboard runs 10 statistics queries  every 5 seconds when a user is on the dashboard, so I figured it was fast enough.

I can update the ADR to include a redis implementation.  Maybe we can centralize that code inside one of the email sending methods (i.e. `send_to_providers.py/send_email_to_provider` to avoid having to track it in 4 places like the sms limit does?

We would also then need to track the number of hard bounces in redis; we'll need to track the synchronous ones in `clients/email/aws_ses.py/send_email()` and the async ones in the `process-ses-result` celery task.  Make sense?

records/2023-01-26.bounce-rate-automated-emergency-brake.md
andrewleith: I was thinking that 10% and greater was a very-bad-thing:tm:, but I will take your recommendation on this :)

records/2023-01-26.bounce-rate-automated-emergency-brake.md
andrewleith: Either value works! As long as `bounce-rate-warning-notification-volume-minimum` isn't less than `bounce-rate-display-volume-minimum` because in that case we would be warning people via but not showing it to them on their dashboard!

I thought I had understood one of the concerns here was notifying many low-volume services and having increased support calls because of it.  The idea was that this value is a lever we can use to reduce those support calls.  

That said, looking at the data over the past few days, a value of 1000 would mean watching 11 services vs. a value of 2500 means watching 6; not a huge difference.  I'll update this as per your recommendation, and we can always tweak this as more services onboard.

jimleroyer: Yes you are right, we would target more services and these would get into our support/education scope, hence more work. If we select a value that is too high though, the risk is that a high bounce rate service excluded from a high individual service limit could influence the global bounce rate. The 2,500 value seems low enough to avoid that, but I like to be extra safe with `1,000` on that front.. but I'd be fine with 2,500 too! It's a type 2 decisions anyway so we'd be fine.

records/2023-01-26.bounce-rate-automated-emergency-brake.md
jimleroyer: My thinking is that one service bounce rate of 10% will not get past the global bounce rate past 10% because the other services will amortize the latter. It could be incorrect if only one service sends notifications for a long enough period of time which is unlikely, especially considering that GCNotify constantly sends notifications for heartbeat (for example).

andrewleith: Yes this is a good line of thinking and a good assumption!

jimleroyer: My last comment made me thinking though: would it be possible for GCNotify to get past 10% bounce rate? Could there be a possibility where we'd suspend our own service? We might want to take explicit conditions to take out GCNotify from such automatic suspension and rather raise critical alarms if that every happened. Let's add this to the ADR please. 🙏

records/2023-01-26.bounce-rate-automated-emergency-brake.md
jimleroyer: I'd avoid repetitive reads on the database if possible. There were cases where we brought down the admin UI because of the DB repetitive reads. I tried to find the old bug card about it (in Trello), but in general that would be a fitting task for a cache server.

andrewleith: Sounds good.  As an aside, we may want to look into the dashboard because as I mentioned it does multiple queries every 5 seconds whenever someone has a dashboard open.

jimleroyer (approved):
LGTM for a merge! we can always update further. I would let the rest of the team review though such as @yaelberger-commits , because there are some products description in the ADR.

PR #41, opened 2023-05-02
Updating secrets usage ADR

jimleroyer:
# Summary | Résumé

Updated the secrets usage ADR for the common secret and salt we use throughout the application with latest changes that occurred, i.e. latest secrets rotation 2023 edition.

Possible yes, we did run some of the changes by the SRE/security team to review. We can ping them again to confirm the changes we actually applied.


records/2022-04-14.secrets-usage.md
```diff
@@ -109,37 +109,37 @@ that are not available through the usual API offered to Notify users.
...
+| Password reset code / Verification URL / Confirmation URL / Invitation URL | Yes        | Yes            | Expires after 1 hour.                                                                                                           | 1. [url_with_tokens](https://github.com/cds-snc/notification-api/blob/bdf1174440272f3bc6d05587cb9100edbe12e3a9/app/utils.py#L29-L34)                                                                                                                                                                                                                                                                                                                                                                                                                                                        |   |
```
jimleroyer, line 125: These URLs still use the salt per he code review I have done. @sastels I assume you might prefer a contextual salt for these as well, right?

sastels: So the app uses the utils repo to generate these. I've added code to utils to use the salt "token" as a default but currently api is still sending in DANGEROUS_SALT. Working on a slight tweak to utils today to make "token" used for creating the token and (either) "token" or DANGEROUS_SALT used for verifying. Once this is merged and the codebase is using it for a day, all the old DANGEROUS_SALT tokens will have expired and we can switch to verifying with only "token".

So TL;DR What you have here is correct and can be merged in. We can update next week when we're not using DANGEROUS_SALT at all for the tokens.

sastels (approved):
LGTM!

records/2022-04-14.secrets-usage.md
```diff
@@ -123,7 +123,7 @@ table below.
...
+| Password reset code / Verification URL / Confirmation URL / Invitation URL | Yes        | No            | Expires after 1 hour.                                                                                                           | 1. [url_with_tokens](https://github.com/cds-snc/notification-api/blob/bdf1174440272f3bc6d05587cb9100edbe12e3a9/app/utils.py#L29-L34)                                                                                                                                                                                                                                                                                                                                                                                                                                                        |   |
```
sastels, line 126: Just noticed the "Expires after 1 hour" - this is true for everything except the invitation, which expires in 2 days.

jimleroyer: Changed it!

PR #43, opened 2023-05-17
Secrets details ADR

sastels:
# Summary | Résumé

Some details of the purpose and usage of Notify secrets.

There is no information in this ADR that could compromise Notify security.


andrewleith (commented):
Looks good to me overall, a few editorial comments below.

PR #45, opened 2023-06-29
Consolidate environment variables

ben851:
# Summary | Résumé

First crack at env var consolidation ADR

---

No security concerns.

      langues officielles?
      fonctionnalité existante?


jimleroyer:
That looks like a real nice improvement over what we currently have. Let's talk to address some points raised by others and a few questions I have:

1. I am not sure if the diagram has the right arrow directions, that's minor but I could help on this if I have the source diagram.
1. Let's see if we can split between app config values and secrets.
1. We might have to store each secret individually in the store to protect leaks, as previously raised by Calvin.

jzbahrai (approved):
Great job explaining and walking through the plan end to end.

ben851:
> This will be awesome! A couple questions:
> 
> 1. To modify variables, will the process be to fetch the yaml file from Secrets Manager, modify it, and push it back?
You can actually modify it directly in AWS Secrets manager, no need to push/pull

> 2. Should we instead encrypt this yaml file and store in github as we do for the k8s secrets (and have the merge action push it to Secret Manager)? This might make review and revert of these secrets easier?
Storing secrets (even encrypted ones) with source code is generally not very good practice, secret versioning will be handled by the AWS secrets manager itself:
https://docs.aws.amazon.com/secretsmanager/latest/userguide/getting-started.html#term_version
We'll have to investigate the best way to review these changes via code reviews etc - and probably manually tag the secrets as we change versions so we can have a complete history. Definitely something to look into!

ben851:
> That looks like a real nice improvement over what we currently have. Let's talk to address some points raised by others and a few questions I have:
> 
> 1. I am not sure if the diagram has the right arrow directions, that's minor but I could help on this if I have the source diagram.

Sure lets touch base and review!

> 2. Let's see if we can split between app config values and secrets.

While this is certainly possible, I struggle to see the benefit of doing so in a devops team... It will add to the complexity of the setup (two files to manage instead of one), and there's no one on the team who won't have access to both files. I'll definitely put my thinking cap on here and see what I can do. 

> 3. We might have to store each secret individually in the store to protect leaks, as previously raised by Calvin.

This has basically the same leak vulnerability as our current solution of lastpass TF Vars, so I'm not sure this is 100% necessary (also see points above). Splitting to separate secrets is probably a good best practice but won't really work in this case... To go that way, we'd probably have to write some custom code to prompt the user to insert each value individually so that it could create the multiple secrets.

records/2023-06-28.environment-variable-consolidation.md
```diff
@@ -0,0 +1,179 @@
...
+
+Since the changes here are not to variables themselves but instead on how they are read into terragrunt, migration is relatively easy and straight forward:
+
+1. Create a new yaml file based off of the example
+2. Copy all variable values from the TF Vars file in LastPass to  the appropriate sections in the yaml file
+3. Copy all variable values set in the individual terragrunt.hcl files to the appropriate sections in the yaml file
+4. Remove all references to variables that aren't dependencies in the terragrunt.hcl files
+5. Copy all variable values set in the env_vars.hcl file to the appropriate sections in the yaml file
+6. Remove the env_vars.hcl file
+7. Modify the Github Actions pipeline to pull the secrets from aws secret manager (a shell script will have to be written)
```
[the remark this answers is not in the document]
ben851: Yep exactly! That, and the file will be a yaml not tfvars. So the procedure would be:
1. fetch environment.yaml from secrets manager
1. in terminal, export TG_VAR_FILE=environment.json
1. terragrunt plan

records/2023-06-28.environment-variable-consolidation.md
```diff
@@ -0,0 +1,179 @@
...
+
+When complete, terragrung plan should show no changes to the infrastructure since the values remain unchanged.
```
ben851, line 172:
```suggestion
When complete, terragrunt plan should show no changes to the infrastructure since the values remain unchanged.
```

jimleroyer:
>>Let's see if we can split between app config values and secrets.

>While this is certainly possible, I struggle to see the benefit of doing so in a devops team... It will add to the complexity of the setup (two files to manage instead of one), and there's no one on the team who won't have access to both files. I'll definitely put my thinking cap on here and see what I can do.

At a process level, treating these two with a slightly different process could reduce the likelihood of errors with regards to their nature, especially while handling secrets. This can reflect at the technical level as well, where it is beneficial to treat secrets different form app config because as previously mentioned by Calvin, we could save each secret individually as that would provide proper labeling and prevent potential secrets leak. We could save the application configuration in only one big file.

Does that make more sense?

>> We might have to store each secret individually in the store to protect leaks, as previously raised by Calvin.

> This has basically the same leak vulnerability as our current solution of lastpass TF Vars, so I'm not sure this is 100% necessary (also see points above). Splitting to separate secrets is probably a good best practice but won't really work in this case... To go that way, we'd probably have to write some custom code to prompt the user to insert each value individually so that it could create the multiple secrets.

Ok let's touch base with @CalvinRodo to validate/invalidate this assumption.

CalvinRodo:
Yes ideally configuration and secrets would be handled different, but at the moment with they way Notify operates we can't really implement least privilege because everyone that needs to debug prod needs admin access to the prod AWS Accounts. 

It's a thing to strive for, but not a thing to prevent implementation of this solution.

CalvinRodo:
So from a risk perspective, storing them seperately could reduce the risk of compromise and aligns closer to Github Secret Best practices, however we still bring it all together when a plan/apply is run and that's in my opinion the greatest chance of it being compromised. 

I think at rest in AWS it's fairly safe to store together unless our AWS is compromised and if that happens then all the secrets are at risk as well as the running code. The benefits in AWS are finely grained access (which doesn't really work for our operational context) and better auditing. 

From a github perspective their best practice recommends not storing secrets in a single structured document, mainly because it increases the risk of secrets being printed to logs. However maybe registering the entire structured data plus iterating over all secrets and registering them as secrets may prevent this risk, we'll probably want to do some POCs to test this out. 

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets

ben851:
> At a process level, treating these two with a slightly different process could reduce the likelihood of errors with regards to their nature, especially while handling secrets. This can reflect at the technical level as well, where it is beneficial to treat secrets different form app config because as previously mentioned by Calvin, we could save each secret individually as that would provide proper labeling and prevent potential secrets leak. We could save the application configuration in only one big file.

Yeah ok, I didn't put too much thought into the process level of things - and for sure that makes sense. I'll do some more thinking on this and try to come up with a happy compromise.

records/2023-06-28.environment-variable-consolidation.md
```diff
@@ -0,0 +1,201 @@
+# Terraform Environment Variable Consolidation
+
+Date: 2023-06-28
+
+## Status
+
+**DRAFT**.
```
jimleroyer, line 7:
```suggestion
**APPROVED**.
```

PR #49, opened 2023-11-21
QuickSight

sastels:
# Summary | Résumé

ADR for AWS QuickSight

no.


records/2023-11-21.quicksight.md
```diff
@@ -0,0 +1,56 @@
...
+
+### QuickSight
+
+QuickSight is an AWS product that we can run in ca-central-1. We can use terraform to create and modify AWS Quicksight subscriptions in our accounts.
+
+### Data
+
+QuickSight imports data from the Notify database into optimized data storage called SPICE. This allows queries to run faster and without creating a load on RDS. SPICE datasets (ie tables) are updated either manually or on schedule. Updates can either be full or incremental.
+
+We specify the data fields added to the QuickSight datasets. We are excluding any fields containing PII or user data (for example, email addresses, phone numbers, and template personalisation).
```
andrewleith, line 30: While I understand why we would do this (and I'm not really suggesting we shouldn't), do you think it will hamper our analysis efforts at all?  Will anyone outside of CDS have access to quicksight?

CalvinRodo: Employees of CDS should not have access to the protected data of the users of our system, I think we would have to have very strong justifications and access controls in place before we considered including this data in quicksight.

records/2023-11-21.quicksight.md
```diff
@@ -0,0 +1,56 @@
...
+
+QuickSight is outside our VPC and requires a AWS QuickSight VPC connection to access our database.
+
+### User access
+
+QuickSight can be configured so that AWS SSO is needed to access the QuickSight subscription. To access the data sets users must also be added one of the QuickSight user groups `dataset-owners` or `dataset-viewers`. Owners may change datasets or add new datasets through the AWS Console (though we will likely do this through terraform). Viewers can only view datasets and use them to create analyses.
```
CalvinRodo, line 36: I would say that connecting to Quicksight through AWS SSO is 100% not negotiable, this allows us to tie into our existing access control around cloud, and also allows us to put restrictions like must authenticate using a yubikey, which also should be required for folks accessing quicksight.

records/2023-11-21.quicksight.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Auditing
+
+QuickSight can be configured to log audit data to CloudTrail.
```
CalvinRodo, line 40: So without PII this is less critical but it should be enabled in case PII slips through, and just as a general CYA for us.

records/2023-11-21.quicksight.md
andrewleith: I generally agree, I guess why I'm asking is because the "to" email address is considered PII but we can probably get a lot of insights from it.  Maybe we could just show the domain rather than the full email address or something? Know what I mean?

jimleroyer: Yeah there are some queries that are quite useful around domains as you mention @andrewleith . We can perform transformations from the DB to Quicksight, hence we'll finetune things as we learn. 

It's an area too where a OLAP schema with fact tables would be best, as we're losing some valuable information when we archive the emails. We could transform the data into domains to remove PIIs and dump summaries in there. We could do it now inside our current schema as this is where we host fact tables at the moment.

records/2023-11-21.quicksight.md
jimleroyer: Yes. I also asked @sastels if we could get more sophisticated roles in this ADR, past the admin, authors and reader, one more aligned with business needs. For example, should growth team requires the same as the support team? Maybe yes, or not, but let's document it in this ADR.

records/2023-11-21.quicksight.md
```diff
@@ -0,0 +1,90 @@
...
+- $0.38/GB per month for extra SPICE (each admin or author account includes 10 GB of free SPICE)
+
+For production we can roughly estimate:
+- 10 admins or authors (total)
+- 10 readers
+- SPICE requirements approximately 80 GB initially
+
+The total cost in this case would be $290/month (no additional SPICE purchase required)
+
+Note that the October 2023 cost for Staging QuickSight was $72 (for 3 users).
```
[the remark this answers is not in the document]
jzbahrai: One thing I am concerned about is if we want to think of this tool as one that everyone can use (a replacement for blazer for all of cds to get access to) - is the cost going to be prohibitive? 72$ for 3 users is quite a bit.

records/2023-11-21.quicksight.md
sastels: Indeed, we can add permissions on more granular levels as desired (ex: different permissions for certain datasets or columns of datasets)

records/2023-11-21.quicksight.md
sastels: 👍 As we discovered this is enabled by default. 🎉

records/2023-11-21.quicksight.md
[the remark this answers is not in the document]
sastels: indeed, it's $24 / author per month, which will add up quickly. We should see who is authoring dashboards / analyses and who just needs reader access.

records/2023-11-21.quicksight.md
sastels: For the V1 of QuickSight I think we should restrict to no PII, using Blazer for any PII related queries.

jzbahrai (approved):

records/2023-11-21.quicksight.md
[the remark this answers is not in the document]
jimleroyer: @sastels I see that was already added in there. Thanks.

PR #50, opened 2023-11-27
System status

andrewleith:
# Summary | Résumé

ADR to lay out how we might implement the system status site.


records/2023-11-24.system-status.md
```diff
@@ -0,0 +1,93 @@
...
+### Back-end
+A new back-end API wil be deployed that will be responsible for assessing the status of each component of the system.  
+
+#### Status checks
+- The website/API/File upload checks can all be done via simple HTTP requests to the respective systems.  No special permissions or rights are required. 
+    - **Question**: Can we leverage [the CDS status page](https://github.com/cds-snc/status-statut) for this?
+- For email and SMS, we will need to assess the actual sending time for each priority of each message type (email and SMS).  These checks will require read access to the notify database in order to calculate sending time.
+   - The healthcheck lambda should be extended to send a notification of each type and priority, and we can use these to assess the health of the email and sms sending functions.
+
+## Code architecture
```
jzbahrai, line 72: Another option would be to release a single admin pod to a standalone kube cluster. Though it would have to hit admin/api in the original eks cluster.

records/2023-11-24.system-status.md
```diff
@@ -0,0 +1,93 @@
...
+A new front-end website will be deployed to show the status information.  It will be served from a new domain, status.notification.canada.ca
+
+### Back-end
+A new back-end API wil be deployed that will be responsible for assessing the status of each component of the system.  
+
+#### Status checks
+- The website/API/File upload checks can all be done via simple HTTP requests to the respective systems.  No special permissions or rights are required. 
+    - **Question**: Can we leverage [the CDS status page](https://github.com/cds-snc/status-statut) for this?
+- For email and SMS, we will need to assess the actual sending time for each priority of each message type (email and SMS).  These checks will require read access to the notify database in order to calculate sending time.
+   - The healthcheck lambda should be extended to send a notification of each type and priority, and we can use these to assess the health of the email and sms sending functions.
```
jzbahrai, line 70: can the above two also live in the lambda? Should they live in the same lambda or should we make separate ones for each request (Im thinking latter)

records/2023-11-24.system-status.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Code architecture
+In order to maintain good developer experience, reduce code duplication and decrease maintenance, the code for the new site should be co-located in the admin repo. We may be able to use [this poetry plugin](https://pypi.org/project/poetry-multiproject-plugin/) to achieve this.  The new pages would not get deployed with admin, but rather be deployed as their own stand-alone site.
+
```
whabanks, line 74: Should we add some implementation details per-component, after discussion of course, on how we'll approach monitoring each component here, or do we want to keep it high level?

andrewleith: The "Proposed component status definitions" table lays out how the statuses will be determined, but you must mean something else?

whabanks: Each component is outlined nicely in the `Status Checks` section but is fairly high level. We can add detail later, since we need to discuss this first but considering we're creating a new project for this it'd be nice to spend some time detailing a proposed project structure and per component implementation details.

Things like:
- What do we want Admin's project structure to look like? Assuming the Poetry plugin is a viable solution.
- Will the project call the API or have it's own independent DB connection?
- How we'll extend the health-check lambda

Maybe I'm missing some context or getting ahead of myself and this is the plan all along 😅.

jimleroyer:
That looks exciting, to have all these status checks available to our users.

Did we analyze the feasibility of also providing a status check on the receipts? to also cover the part that notifications actually reached mailboxes or SMS, on top of sending these?

records/2023-11-24.system-status.md
```diff
@@ -0,0 +1,162 @@
...
+1. API-lambda
+    We can change the current heartbeat API lambda to make all the API calls required for the system status page.
+1. ECS Cluster
+    We can set up an ECS task (similar to Blazer/Perf-tests) to allow all the API calls
+
+Given we use both versions in our current code base and neither has any significant advantage over the other, we are going to use the API lambda (Option 1)
+
+The API lambda needs to accomplish the following things:
+1. Call the different APIs to get the information to display on the status page
+1. Call the Articles endpoint to get information around Incidents
```
andrewleith, line 79: @jumana, I think we could just do this from the front-end instead, and not have to copy/relay the content.

yaelberger-commits (approved):

records/2023-11-24.system-status.md
andrewleith: These are great questions!

> What do we want Admin's project structure to look like? Assuming the Poetry plugin is a viable solution.

We decided after the last meeting that we don't need to keep the same look and feel for both, we won't have to try to tie these together, which in the end will make deployment much easier.  I'll update the ADR to reflect this.

---
> Will the project call the API or have it's own independent DB connection?

It would call the DB itself as we wouldn't want to rely on API for this so that we can report the distinction when API is down but celery is still chugging along sending notifications.

---
> How we'll extend the health-check lambda

I have a one-liner on this, not sure if we need more detail or not:
> The healthcheck lambda should be extended to send a notification of each type and priority, and we can use these to assess the health of the email and sms sending functions.

PR #52, opened 2023-12-11
System status ADR update

andrewleith:
# Summary | Résumé

- Add info about using s3/cloudfront instead of github pages.
- Add info about AWS outages
- Update component status definitions
- Update where json is written (s3 and not github)

### DIFF
--- a/records/2023-11-24.system-status.md
+++ b/records/2023-11-24.system-status.md
@@ -27,24 +27,24 @@ Each system component can be rated as one of the following statuses:
 
 Each component will have its own criteria for determining its status, based on our [SLO](https://notification.canada.ca/service-level-objectives) where possible.
 
-### Proposed component status definitions
-| Component   | Status   | Definition                                                                             |
-| ----------- | -------- | -------------------------------------------------------------------------------------- |
-| Website     | UP       | The website is responding to requests in 400ms or less                                 |
-| Website     | DEGRADED | The website is responding to requests in more than 400ms                               |
-| Website     | DOWN     | The website is not responding                                                          |
-|             |          |                                                                                        |
-| API         | UP       | The API is responding to requests in 400ms or less                                     |
-| API         | DEGRADED | The API is responding to requests in more than 400ms                                   |
-| API         | DOWN     | The API is not responding                                                              |
-|             |          |                                                                                        |
-| Email       | UP       | Priority emails are sending within 60s, and Other emails are sending within 45m        |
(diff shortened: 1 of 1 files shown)


records/2023-11-24.system-status.md
```diff
@@ -83,18 +83,12 @@ The API Lambda will interact with 2 different systems:
 We will write all the above results into a singular .json and commit it to github.
```
jzbahrai, line 83: does this need to be changed? as we are now writing the .json to an s3 bucket?

jzbahrai: line #75 as well

records/2023-11-24.system-status.md
```diff
@@ -83,18 +83,12 @@ The API Lambda will interact with 2 different systems:
...
-Github can host static webpages.
-The github repo will display status.notification.canada.ca
+### Front-end site
+We will use s3+cloudfront to serve up both staging and production versions of the site (`status.staging.notification.cdssandbox.xyz` and `status.notification.canada.ca`)
 
 #### The final flow will look like
 
-API Lambda -> Github Repo -> Static Page
-
-### Considerations
```
jzbahrai, line 94: we should leave this consideration - we are still deploying within AWS and have the issue of being on the same cloud provider having issues... aka we have increased our risk

records/2023-11-24.system-status.md
andrewleith: We could still write/read the json file from github.. I hadn't intended to change that, but perhaps we should for resiliency sake?

records/2023-11-24.system-status.md
CalvinRodo: AWS Wide outages are very rare, more like a single region would go down, so cross region replication would be probably a good enough mitigation for that issue. 

Just using CloudFront should be enough, as it does edge caching so it should be very low risk of it going down.

If you really want a backup we can always deploy to azure and host the static site there instead or as well so we have a backup.

CalvinRodo: If you are curious here is a list of all the major outages AWS has had since it's inception, s3 going down is 100% considered a major outage due to how important it is to AWS

https://awsmaniac.com/aws-outages/

andrewleith: Yeah I think this covers us plenty, much better than our GH pages plan!

records/2023-11-24.system-status.md
andrewleith: I will update this to use S3 as well.

jzbahrai (approved):

PR #54, opened 2024-01-22
Integrating the dev environment 

jimleroyer:
# Summary | Résumé

ADR for integrating the new dev environment into our CI/CD and potentially automating furthermore the overall release process.

### DIFF
--- a/records/2024-01-22.dev-environment-cicd.md
+++ b/records/2024-01-22.dev-environment-cicd.md
@@ -1,0 +1,73 @@
+# TITLE
+
+Date: 2024-01-22
+
+## Status
+
+**DRAFT**.
+
+## Context
+
+A new *development* environment has been created with recent effort around
+business continuity plan (BCP)<sup>[1]</sup>. It was created from scratch
+with the goal of determining how much effort and time it takes to recreate
+a whole GCNotify environment from scratch. Since then, the team used this
+environment to experiment on new infrastructure changes in an isolated manner,
+increasing velocity with changes of this nature.
+
+This also comes at a time when there is a desire to automate furthermore the
+release process among the team, potentially fastening the release process and
+decreasing the time spent on the ops lead routine tasks. This has to be done
+with certain considerations in mind such as risk, testing, traceability,
+rollback and accountability.
+
+> <sup>[1]</sup> A business continuity plan (BCP) is a strategic playbook created to help an organization maintain or quickly resume business functions in the face of disruption, whether that disruption is caused by a natural disaster, civic unrest, cyberattack, or any other threat to business operations.
+
+## Goals
+
+It is necessary to take a step back and decide what are the team's primary's objectives for this effort.
+
+### More automation
+
+_TODO: fill in the blanks_
+
+### Less chores for the ops lead
+
+_TODO: fill in the blanks_
+
+### More release autonomy
+
+_TODO: fill in the blanks_
+
+### Isolated development environment
+
+_TODO: fill in the blanks_
+
+### Isolated experimental infrastructure environment
+
+_TODO: fill in the blanks_
+
+## Options
+
+This section lists suggestions made by the team with different considerations
+and goals. We'll first review the current CI/CD process.
+
+### Current process
+
+_TODO: Insert current process diagram._
+
+### Option 1
+
+_TODO: Describe an option._
+
+## Additional considerations
+
+_TODO: Describe extras in here._
+
+## Decision
+
+_TODO: Describe the way forward._
+
+## Consequences
+
+_TODO: Describe foreseen and felt consequences of the decision (possible after 1-3 months)._


records/2024-01-22.reshaping-cicd-environments.md
```diff
@@ -0,0 +1,127 @@
...
+environment.
+
+## Options
+
+This section lists suggestions made by the team with different considerations
+and goals. We'll first review the current CI/CD process.
+
+### Current process
+
+![Current CI/CD workflow](https://raw.githubusercontent.com/cds-snc/notification-adr/main/records/diagrams/current-cicd.drawio.png?raw=true "Current CI/CD workflow")
```
jzbahrai, line 73: this doesn't seem to be linked properly:/ and doesn't show up when I am viewing in markdown

records/2024-01-22.reshaping-cicd-environments.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Stronger responsibility for release changes on developers
+
+A lightweight changes that the process can include in the first place is to put
+more burden on the developers' shoulders on the changes approval step.
+Specifically, whatever the developer pushes to the staging environment is what
```
jzbahrai, line 79: does it get released immediately? or is there a time period before the push to prod happens?

records/2024-01-22.reshaping-cicd-environments.md
```diff
@@ -0,0 +1,127 @@
...
+- Very cheap to implement
+- Lighten load on the ops lead role
+
+#### Cons
+
+- Higher risk factor of deploying to production (as it relies on multiple human interventions)
+
+### Create dedicated "Core" and "Dev" environments
+
+Currently the staging environment is used as an integration/development environment. This has been workng, but is not aligned with best practices, as it leaves staging in a non-production like state. With the introduction of the new "Dev" environment, it became possible to add an additional environment into the process that allows staging to always remain in the same state as production.
```
jzbahrai, line 98: Can we add when it hasn't been working? Right now the only thing we test on staging is callbacks I believe.

records/2024-01-22.reshaping-cicd-environments.md
```diff
@@ -0,0 +1,127 @@
...
+
+Currently the staging environment is used as an integration/development environment. This has been workng, but is not aligned with best practices, as it leaves staging in a non-production like state. With the introduction of the new "Dev" environment, it became possible to add an additional environment into the process that allows staging to always remain in the same state as production. 
+
+The new "Dev" envorinment as it exists today is not suitable for such a role because it is constantly being created/destroyed/altered by the Core team in order to test and develop infrastructure changes. It would not be reliable enough for the App team to develop against. 
+
+An additional environment could be created that is reserved for the App team's development and integration, and the existing "Dev" environment could be reclassified as the "Core" environment to identify it as being used by the Core team for infrastructure development.
+
+![Potential Core and Dev Environment Solution](https://raw.githubusercontent.com/cds-snc/notification-adr/main/records/diagrams/2024-01-22.reshaping-cicd-environments/core-dev-env.png?raw=true "Potential Core and Dev Environment Solution")
```
jzbahrai, line 104: same issue as above - the link is not rendering correctly

records/2024-01-22.reshaping-cicd-environments.md
```diff
@@ -0,0 +1,127 @@
...
+
+#### Pros
+  - Staging could be used to test the actual production release package and the associated process, which is currently not being tested.
```
jzbahrai, line 107: Can you add an example here? Specifically around the part - the associated process, which is currently not being tested. I am unsure what that is

records/2024-01-22.reshaping-cicd-environments.md
```diff
@@ -0,0 +1,127 @@
...
+  - The release process to staging could be automated completely. This would provide a safety net for automating releases to production. I.e: at end of day, automatically create a release package from Dev, and automatically deploy this package to staging. Run automated testing overnight. If the tests pass, there is a high degree of certainty that it will release successfully in production, and thus does not require intervention from the team/Ops Lead.
+  - Application development would not be hindered by the Core team's infrastructure changes
+
+#### Cons
+  - Potentially increased cloud costs, but there are ways to reduce/eliminate this
+  - Increased complexity for Core team with environment management
+  - Increased complexity for release process
+
```
jzbahrai, line 115: - We could in theory tag the release and run performance tests overnight on staging (I think we already do this?). And then release to prd in the morning? vs adding an env overhead.

records/2024-01-22.reshaping-cicd-environments.md
jimleroyer: The link is targeting the final version once the PR is done, but you can see the picture attached to this PR though. If we target the link in the PR, it will break once merging into `main`.

records/2024-01-22.reshaping-cicd-environments.md
jimleroyer: No automation improvement on this one option, but only a slight change to unburden the ops lead.

records/2024-01-22.reshaping-cicd-environments.md
jimleroyer: The link is targeting the final version once the PR is done, but you can see the picture attached to this PR though. If we target the link in the PR, it will break once merging into main.

records/2024-01-22.reshaping-cicd-environments.md
jimleroyer: I am not sure if I understand correctly the question with reference to the callbacks. Can you expand on that? or hit me on Slack?

records/2024-01-22.reshaping-cicd-environments.md
jimleroyer: Per my understanding, we mean integration wise: we test and QA the individual parts but not the whole snapshot. For example, a backend change in the *notification-api* repository could lead to a bug with the front-end in a subtle manner, but we won't notice because we only run front-end tests with the *notification-admin* changes. Same can apply with the performance tests: we could run a subset of these against the release snapshots (but we don't).

Treating staging as dev also puts us in a position where the former environment cannot be released at certain point in time if we'd need to do a hot patch for an emergency release, because dev changes are in and these are not ready or tested.

records/2024-01-22.reshaping-cicd-environments.md
jimleroyer: It's a possibility that I will include as a different option, probably under a release train. Each will have their own set of pros and cons. What would get picked depends on what the team feels comfortable with in terms of risk, complexity, effort, budget, process alignment and automation (if I can summarize the initial goals that way).

jzbahrai:
In regards to Steve's comment on cutting off the release at a reasonable time - we will need to define a process. Nothing should be merged from say 5pm EST - 9/10am EST when the release goes out? Something like that?

jimleroyer:
@jzbahrai I am not sure if @sastels was pushing for a release cut, or distancing main branch from the release (with an additional branch for example, and/or additional env), which can be 2 different strategies. We'd need to talk of the options again in a dev code review meeting.

records/2024-01-22.reshaping-cicd-environments.md
```diff
@@ -0,0 +1,236 @@
...
+for a production release would get merged into the staging environment, most
+likely from a `dev` to the `main` VCS branch transition.
+
+Instead of kicking the release process with a manual action, the release would
+be done with changes promoted and deemed worthy of production, since these
+changes were promoted by task workers independently and quality checks such as
+quality assurance, product manager's approval, designer's checks.
+
+#### Pros
+
```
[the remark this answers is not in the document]
jimleroyer: I will add it. The provided examples are automated already though, right?

PR #64, opened 2024-04-11
Inital work on the ADR for improving 2FA login

whabanks:
# Summary | Résumé

> 1-3 sentence description of the changed you're proposing, including a link to
> a GitHub Issue # or Trello card if applicable.

> Description en 1 à 3 phrases de la modification proposée, avec un lien vers le
> problème (« issue ») GitHub ou la fiche Trello, le cas échéant.

---

      langues officielles?
      fonctionnalité existante?

### DIFF
--- a/records/2024-04-11.2FA-magic-link.md
+++ b/records/2024-04-11.2FA-magic-link.md
@@ -1,0 +1,53 @@
+# Add magic links to 2FA emails
+
+Date: 2024-04-11
+
+## Status
+
+**DRAFT**.
+
+_VALUES: (DRAFT, IN REVIEW, APPROVED, REJECTED)_
+
+## Context
+
+Notify requires users to manually copy two factor authentication codes from their email then paste it into a text box on the site in order to log into Notify. This process is cumbersome and not that accessible, or easy to use.
+
+The aim of this initiative is to improve the accessibility and ease of use of Notify's login process.
+
+## Options
+
+### Increase the visibility of 2FA codes in emails
+The simplest approach is to increase the visiblility of codes in the email to make the code more obvious and easier to interact with:
+- Place the code on it's own line in the email
+- Bold or otherwise highlight the next
+- Increase font size
+
+While increasing visibility is an improvement, it does not address the cumbersome amount of actions needed to complete the login:
+
+```mermaid
+flowchart LR
+	id1(Open email)-->id2(Find the code)-->id3(Highlight it)-->id4(Copy it)-->id5(Switch to Notify tab)-->id6(Select 2FA code text box)-->id7(Paste in the code)-->id8(Click Submit)
+```
+
+### Use magic links
+By using a magic link that the user can simply click on to complete the login process, we reduce the number of manual interactions required by the user to log in, thus making the process less obstructive and feel more natural.
+
+
+```mermaid
+
+flowchart LR
+	id1(Open email)-->id2(Find the magic link)-->id3(Click the link)-->id4(Logged in)
+
+```
+
+## Additional considerations
+
+_TODO: Describe extras in here._
+
+## Decision
+
+_TODO: Describe the way forward._
+
+## Consequences
+
+_TODO: Describe foreseen and felt consequences of the decision (possible after 1-3 months)._


records/2024-04-11.2FA-magic-link.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Email Impersonation
+A malicious actor could send a user, who just logged in, an email containing a magic link of their own. This link could redirect the user to a phishing site enabling the malicious actor to collect partial auth information from the user trying to log in, providing a window where the attacker could obtain unauthorized access to Notify.
+
+The attack surface, while not impossibly small, relies on a number of time limited circumstances to be aligned for this type of attack to succeed:
+
+1. The attacker needs to know the moment the user clicked the sign-in button
+2. The attacker needs to send their phishing email before the legitimate email is received to maximize their success -OR- they would need to execute a man-in-the-middle attack to modify the legitimate email's magic link to point to the malicious site.
+3. If the attacker succeeds in phishing and retrieving partial login information, they would also need access to the real magic link that we sent to the user in order to successfully log in.
```
jimleroyer, line 72: I think there was also a mention that a cookie needs to be present on the client side after username/password login and prior to click on the magic link. This prevents phishing attempt as it would require the attacker to also possess that cookie, or the username/pw along with mail inbox access of the target.

records/2024-04-11.2FA-magic-link.md
```diff
@@ -0,0 +1,135 @@
...
+        id6(Select 2FA code text box)-->id7(Paste in the code)-->id8(Click Submit)
+    end
+    id0 --Tab Switch--> id1
+    id4 --Tab Switch--> id6
+    id8-->id9((X))
+
+```
+
+### Use magic links
+By using a magic link that the user can simply click on to complete the login process, we reduce the number of manual interactions required by the user to log in, thus making the process less obstructive and feel more natural.
```
jimleroyer, line 43: Can we please explain that what we define here as magic links might not be true magic links as generally understood in the industry? i.e. we still require a password prior to sending the link, whereas a magic link solution would only require the email information, if I understood correctly.

records/2024-04-11.2FA-magic-link.md
whabanks: Good point about the cookies. I've re-worked the security considerations section to, hopefully, reflect a more accurate assessment of the threats we discussed with the security team.

records/2024-04-11.2FA-magic-link.md
whabanks: Your understanding is correct, I've added detail on how our implementation would differ from the typical industry standard.

records/2024-04-11.2FA-magic-link.md
```diff
@@ -0,0 +1,171 @@
...
+### Email Phishing
+An attacker could tamper with the 2FA email we send to the user and forward them to a phishing site that prompts them to re-input their username and password. If the user is not aware of this type of risk they could unintentionally provide an attacker with their log in credentials.
+
+In this instance, the attacker now has a means to create their own partially authenticated session cookie, and would only need access to the magic link. If they have successfully tampered with the 2FA email, we can assume the attacker already has the means to obtain the magic link, even without direct access to the user's inbox.
+
+#### Mitigations
+* Provide guidance to inform users of the risks of magic links, and steps they can take to mitigate them.
+* See the [Fallback Mitigation section](#Fallback-Mitigation---An-over-engineered-"last-resort"-solution)
+
+###
```
jimleroyer, line 93:
```suggestion
```

records/2024-04-11.2FA-magic-link.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Fallback Mitigation - An over-engineered "last resort" solution
+
+A fallback mitigation layer to protect against scenarios that are out of our control would be to establish a process for generating a unique visitor fingerprint. This fingerprint could be ephemerally stored for the duration of a magic link's validity period. That fingerprint could then be validated once the user clicks the magic link, and before beginning the final steps of the login process.
+
+The fingerprint could use the user's browser attributes, IP address, HTTPS session keys, etc to form a composite key that could then be hashed to uniquely identify the user. Since the success of the attacker is time dependent, [Argon2d](https://github.com/P-H-C/phc-winner-argon2) could be a good candidate for such a hashing implementation as it uses data-depending memory access to make it resistant to GPU based brute force attacks.
```
jimleroyer, line 98: Did you speak with Andrew about using the `secrets` module too (instead of `SystemRandom`)? Do you know how that would compare?

Also, do we consider lowering the time period to use the link? At the moment, the default is 30 minutes for the code sent via the email.

whabanks: > Did you speak with Andrew about using the secrets module too (instead of SystemRandom)? Do you know how that would compare?

I have not, thank you for the reminder about the `secrets` module. I had forgotten about it but I think it could be applicable here - will follow up with Andrew.

> do we consider lowering the time period to use the link? At the moment, the default is 30 minutes for the code sent via the email.

30 minutes is quite lenient / lax to me, I think reducing this time is a reasonable mitigation that we can consider. I will add a section about this.

records/2024-04-11.2FA-magic-link.md
```diff
@@ -0,0 +1,170 @@
...
+        end
+        alt Valid hash
+            Notify->>User's browser: Log user in
+            Notify->>DB: Delete login event / fingerprint
+        end
+    end
+
+```
+
+Such an approach could allow us to regain control over some scenarios that typically arise from user action and behaviour that is out of our control:
```
jzbahrai, line 144: Thanks for explaining this scenario. The cost consideration is something that we need to take into account. What is industry standard here btw? Magic Links are becoming more common, do most folk implement this in the way suggested above?

Is there a link to the threat modeling exercise that was done for this ADR and if so can it be added here? What is the expectation from our security team on what level of defensiveness we need to implement?

andrewleith: Here is the threat modeling doc: https://docs.google.com/document/d/1sIdrJ_dlkzwQoEVss0SMXjhdrLPYUenFlf3Xm8-amUc/edit

P0NDER0SA (commented):
I have read this and I do not have any comments or recommendations to make at this time...

records/2024-04-11.2FA-magic-link.md
```diff
@@ -0,0 +1,170 @@
...
+* The user's device is compromised
+* The user's email inbox is compromised
+* The user forwards their 2FA link to unauthorized parties
+* The user falls victim to phishing attacks
+
+
+## Considerations
+While a unique fingerprint hashed with Argon2d could be an appropriate fallback solution to thwart these types of attacks there are several considerations that add to the scope of such an implementation:
+
+* Policy and Legal implications of collecting users' browser attributes and / or IP addresses.
```
ben851, line 155: This is a very good point that we should run past policy.

ben851 (commented):
This all makes sense to me - my only other comment is that if we implement magic links, I would also prefer if we keep the option to copy/paste the code... maybe make it a user preference if it's too confusing to have both by default? Magic links open new tabs and thus I lose my history.

records/2024-04-11.2FA-magic-link.md
```diff
@@ -0,0 +1,170 @@
...
+
+* Introduce a process for checking that the `Origin` header matches the target origin when a magic link is followed. As our application is distributed and lives behind a load balancer, making each node aware of it's target origin could prove to be non-trivial task and/or have unintended consequences.
+
+See the [Fallback Mitigation section](#Fallback-Mitigation---An-over-engineered-"last-resort"-solution)
+
+### Email Phishing
+An attacker could tamper with the 2FA email we send to the user and forward them to a phishing site that prompts them to re-input their username and password. If the user is not aware of this type of risk they could unintentionally provide an attacker with their log in credentials.
+
+In this instance, the attacker now has a means to create their own partially authenticated session cookie, and would only need access to the magic link. If they have successfully tampered with the 2FA email, we can assume the attacker already has the means to obtain the magic link, even without direct access to the user's inbox.
+
```
sastels, line 88: After you enter your password and get to a "we've sent you a link to click on" page, we could emphasize that if that link brings them to a page that asks for their password then it's a phishing attempt and they should immediately close the page (and maybe forward the email to us). This might help mitigate the risk.

whabanks: I like this idea, especially if we have the option to make it really stand out. My only worry for our existing users is possibly missing out on the added guidance if we put it on that "we've sent you a link" page. I have personally got into the habit of clicking sign-in and immediately swapping to my inbox to fetch the 2FA code. 
Another option would be to place the guidance directly on the sign-in page for short period of time after release to highlight the changes to the sign in process and a blurb about "staying safe".

records/2024-04-11.2FA-magic-link.md
whabanks: Added the sections addressing the `secrets` module migration and detailing reduction of the 2FA validity period as a potential mitigation.

records/2024-04-11.2FA-magic-link.md
whabanks: I added a link to the thread modelling exercise doc under the `Security Considerations` section.

whabanks:
> This all makes sense to me - my only other comment is that if we implement magic links, I would also prefer if we keep the option to copy/paste the code... maybe make it a user preference if it's too confusing to have both by default? Magic links open new tabs and thus I lose my history.

This is doable I think. We'd need to do some design work to ensure that it does not become confusing. We could make it a user profile option / ask users when they sign up whether they prefer a code or a magic link for their 2FA. This way we can either let them know to check their email for a link, or provide them with the usual text box for the 2FA code to be manually input but not both at any given time.

records/2024-04-11.2FA-magic-link.md
```diff
@@ -0,0 +1,178 @@
...
+* The user's device is compromised
+* The user's email inbox is compromised
+* The user forwards their 2FA link to unauthorized parties
+* The user falls victim to phishing attacks
+
+
+## Considerations
+While a unique fingerprint hashed with Argon2d could be an appropriate fallback solution to thwart these types of attacks there are several considerations that add to the scope of such an implementation:
+
+* Policy and Legal implications of collecting users' browser attributes and / or IP addresses.
```
cgye, line 159: IP address is considered PII. See:

- https://www.canada.ca/en/transparency/privacy.html
- https://www.priv.gc.ca/en/opc-actions-and-decisions/research/explore-privacy-research/2013/ip_201305/

Collection of IP addresses needs to be justified and protected (implies additional security controls). I am not convinced that the use case here (not a core Notify functionality) can be used to justify the collection of IP addresses.

records/2024-04-11.2FA-magic-link.md
```diff
@@ -0,0 +1,178 @@
...
+* CSRF protections buit into Flask, and which are already enabled in our application, configure the `Set-Cookie` header with the `HttpOnly` and `Secure` options, preventing Javascript from reading the contents of the session cookie, and limits the cookie(s) to HTTPS traffic only
+
+* The `Strict-Transport-Security` header is enabled site wide, converting all requests to HTTPS, thus mitigating man-in-the-middle attacks.
+
+* Introduce a process for checking that the `Origin` header matches the target origin when a magic link is followed. As our application is distributed and lives behind a load balancer, making each node aware of it's target origin could prove to be non-trivial task and/or have unintended consequences.
+
+See the [Fallback Mitigation section](#Fallback-Mitigation---An-over-engineered-"last-resort"-solution)
+
+### Email Phishing
+An attacker could tamper with the 2FA email we send to the user and forward them to a phishing site that prompts them to re-input their username and password. If the user is not aware of this type of risk they could unintentionally provide an attacker with their log in credentials.
```
cgye, line 85: IMHO, a more likely scenario (and probably easier to achieve) is to create an e-mail which looks like Notify's 2FA email, but with a malicious link (e.g. reflected XSS). The following measures could mitigate this:

- ensure Notify is not vulnerable to XSS (e.g. via automated vulnerability scanning & patching).
- educating end user on how to identify legitimate (e.g. verifying the from field in e-mail header) e-mails from Notify.

records/2024-04-30.magic-link-security-team-proposal.md
```diff
@@ -0,0 +1,112 @@
...
+* An increase in AWS costs, and system performance, related to running the Argon2d hashing algorithm for each login
+* Level of effort - this solution would likely not be trivial to implement, and/or maintain in the future.
+
+# Concept Overview
+
+PKCE uses a pair of related tokens, the code_verifier and the code_challenge, to secure the exchange:
+* Code Verifier: A high-entropy cryptographic random string generated by the client, which is kept secret and used later to verify the authenticity of the request.
+* Code Challenge: A derived version of the code_verifier, usually created by hashing the code_verifier and sending it over the network (in PKCE, this is sent to the authorization server during the initial request).
+
+# Applying PKCE-like mechanism to Magic Links
```
jimleroyer, line 16: That is a good suggestion in the context of magic links. Our magic link differs a bit though in the sense that the user had to authenticate before and we left a cookie on the browser along with a session ID that both backend and front end should know about (on top of my head -- if I am wrong, tell me). 

Hence I have the feeling that this process achieves the same, i.e. to identify and bound the user' session between front-end and back-end, in the context of password-less magic links where there were no prior contact. Maybe the current implementation we have might be improved with the suggested code challenge, I have the impression though our workflow could be different given we don't have "true" password-less magic link.

Does that make sense? Are some of my assumptions wrong? And if we implement this logic for magic links, should we adapt it to our password present workflow and how?

andrewleith: I tend to think you are right here, @jimleroyer.  We achieve basically the same thing with the cookie that is provided once the user provides their username and password.  

Given that is the case, would this be necessary?

jzbahrai:
Moving this to draft as we have decided not to go ahead with this for now

PR #79, opened 2025-02-27
Different 2FA methods

jzbahrai:
# Summary | Résumé

This is the first draft of an ADR on the different 2FA methods available for notify.

---

      langues officielles?
      fonctionnalité existante?

### DIFF
--- a/records/2025-02-19.authmethods.md
+++ b/records/2025-02-19.authmethods.md
@@ -13,7 +13,7 @@ Two-factor authentication (2FA) provides an additional layer of security beyond 
 Notify currently has three authentication methods:  
 1. Email  
 1. SMS  
-1. Yubi-key
+1. FIDO2 supported devices (such as YubiKey)
 
 A client reached out to our support team indicating that their department was going to sunset giving each employee a phone. Employees would be expected to use MS Team phone numbers as the default method for cellular contact. Both the client and the Notify team tested sending an SMS to an MS team number and were unable to do so. Using MS Team numbers for 2FA is not an option.
 
--- a/records/2025-02-19.authmethods.md
+++ b/records/2025-02-19.authmethods.md
@@ -129,6 +129,8 @@ On the profile page, a user should be able to choose any of the recommended auth
 3. Magic link
 4. Email code auth
 
+Recognizing that the second option is mobile-based(see section below for Desktop based Authenticator apps), we are still going to offer authenticator apps for clients who have smart devices.
+
 **If a client doesn't have a yubi key or a phone, we will default to email_auth for them.**
 
 ## Other options that we can revisit in the future
@@ -161,12 +163,12 @@ The GC Notify team should recommend one of the above apps and also provide instr
 
 #### Recommended Implementations
 
-KeePassXC: Open-source password manager with integrated TOTP support
-Authy Desktop: Cross-platform solution with encrypted cloud backup
-2FAS Desktop: Lightweight, security-focused TOTP application
+KeePassXC: Open-source password manager with integrated TOTP support  
+Authy Desktop: Cross-platform solution with encrypted cloud backup  
+2FAS Desktop: Lightweight, security-focused TOTP application  
 
 The GC Notify team should recommend one of the above apps and also provide instructions on best practices.
 
-### Conclusion
+#### Conclusion
 Desktop TOTP applications are an option over mobile-based authenticator apps. They come with their own unique set of challenges that GC Notify would have to provide guidance for.
 
--- a/records/2025-02-19.authmethods.md
+++ b/records/2025-02-19.authmethods.md
@@ -1,0 +1,172 @@
+# Different Authentication Methods for Notify
+
+Date: 2025-02-19
+
+## Status
+
+**DRAFT**.
+
+## Context
+Two-factor authentication (2FA) provides an additional layer of security beyond passwords by requiring a second verification method. This ADR evaluates Notify's current 2FA methods and the future state for 2FA within our system.
+
+## Problem
+Notify currently has three authentication methods:  
+1. Email  
+1. SMS  
+1. Yubi-key
+
+A client reached out to our support team indicating that their department was going to sunset giving each employee a phone. Employees would be expected to use MS Team phone numbers as the default method for cellular contact. Both the client and the Notify team tested sending an SMS to an MS team number and were unable to do so. Using MS Team numbers for 2FA is not an option.
+
+Given the state of Notify's authentication and sign-up flow, this ADR aims to discuss the different auth types for Notify and decide what our auth flow should be.
+
+Discussions about changes to the sign-up flow will follow the acceptance of this ADR.
+
+## Goals
+The goal of this ADR is to identify which authentication methods we should keep using/ if we should implement new ones. 
+
+## Analysis of current auth types
+| auth_type  | count |
+| ---------- | ----- |
+| email_auth | 3,413 |
+| sms_auth   | 3,100 |
+
+Even though GC Notify supports other auth types - yubi key for example, we don't store this in the user profile. This is something that we should fix going forward.
+
+| Current Methods | Pros                                                                                                                                                                                                                                          | Cons                                                                                                                                                                                        |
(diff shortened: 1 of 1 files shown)


records/2025-02-19.authmethods.md
```diff
@@ -0,0 +1,134 @@
...
+This is just an additional 2FA method for clients with a department-provided cellphone.
+
+### Benefits:  
+• Works offline once configured  
+• Not vulnerable to SIM swapping  
+• No SMS costs  
+• Higher security than SMS  
+
+### Cons:  
+• Requires smartphone and app installation
```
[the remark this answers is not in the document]
jzbahrai: this is good idea! I think its definitely an option - I can add this in as a potential option.

records/2025-02-19.authmethods.md
```diff
@@ -0,0 +1,174 @@
...
+**DRAFT**.
+
+## Context
+Two-factor authentication (2FA) provides an additional layer of security beyond passwords by requiring a second verification method. This ADR evaluates Notify's current 2FA methods and the future state for 2FA within our system.
+
+## Problem
+Notify currently has three authentication methods:  
+1. Email  
+1. SMS  
+1. Yubi-key
```
jimleroyer, line 16:
```suggestion
1. FIDO2 supported devices (such as YubiKey)
```

records/2025-02-19.authmethods.md
```diff
@@ -0,0 +1,174 @@
...
+1. YubiKey
+2. Authenticator app
+3. Magic link
+4. Email code auth
+
+Recognizing that the second option is mobile-based(see section below for Desktop based Authenticator apps), we are still going to offer authenticator apps for clients who have smart devices.
+
+**If a client doesn't have a yubi key or a phone, we will default to email_auth for them.**
+
+## Other options that we can revisit in the future
```
jimleroyer, line 136: I would just move these into the options section (and mark these as rejected), or get rid of the section from this ADR to reduce noise.

records/2025-02-19.authmethods.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Passkeys
```
jimleroyer, line 138: There could be a section on passwordless authentication, which YukiKeys are part of as well, but it is better to dedicate such topic into its own scoped ADR.

records/2025-02-19.authmethods.md
```diff
@@ -0,0 +1,174 @@
...
+## 1. Remove SMS as a 2FA option  
+
+Currently, Notify requires a user's phone number in the signup flow. 
+1. We can remove SMS as an option going forward.  
+2. We will need a migration plan to move existing SMS 2FA users to another auth_type.  
+3. We would then disable 2FA verification by SMS for all Notify users.  
+
+Benefits of removing SMS-based 2FA:  
+• Eliminates vulnerabilities to SIM swapping attacks  
+• SMS is known to be the least secure form of 2FA
```
jimleroyer, line 85: It is certainly a strong recommendation now and a trend to remove SMS. While I agree overall on removing SMS as an authentication mechanism, my understanding is there might be a nuance within GCNotify and government emails where we are not at such high risk. The 2 main reasons are because 1) we use 2FA along with the SMS authentication (a SIM swap is not enough), and 2) GCNotify does not make use of SMS to recover the accounts (IIRC), but only via emails. 

That last point is important because that makes SMS less dangerous as a factor authentication: it cannot bypass the email for account recovery. If that was the case, one could do a SIM swap and bypass a government email, and take over the account (but that isn't the case here).

That brings a yellow flag on our current process though: if the email is compromised, then there is no extra factor authentication to prevent the account takeover. This is where normally another mechanism is helpful... but not with SMS obviously but rather a YubiKey or authenticator application.

I mention it because that might play in consideration for a migration effort and timeline to move away from it.

records/2025-02-19.authmethods.md
```diff
@@ -0,0 +1,174 @@
...
+
+## 4. Default to using email for 2FA
+If we remove SMS as a 2FA method, we should default the 2FA to email authentication.  
+• Requires minimal changes to authentication framework as this is already implemented  
+• Verification logic remains largely unchanged from SMS implementation  
+
+# Decision
+
+We are not going to offer SMS 2FA auth going forward. There are two reasons for this:  
+1. SMS 2FA is not secure (as shown in recommendation 1)
```
jimleroyer, line 122: I would add some nuance to this as per my previous comment on recommendation 1.

records/2025-02-19.authmethods.md
```diff
@@ -0,0 +1,174 @@
...
+2. Clients without a phone should still be able to sign up for notify
```
jimleroyer, line 123: This is per my understanding the main issue that was raised to us, and that this ADR needs to solve as per the opening problem. But I don't see why the removal of SMS as a factor authentication is required. 

I am unsure but the only blocker there might be is that we make phone numbers a required part of our registration?  If we remove that requirement, users could then opt for SMS or email for their preferred 2nd factor.

records/2025-02-19.authmethods.md
```diff
@@ -0,0 +1,174 @@
...
+1. We can remove SMS as an option going forward.  
+2. We will need a migration plan to move existing SMS 2FA users to another auth_type.  
+3. We would then disable 2FA verification by SMS for all Notify users.  
+
+Benefits of removing SMS-based 2FA:  
+• Eliminates vulnerabilities to SIM swapping attacks  
+• SMS is known to be the least secure form of 2FA
+• Improves overall security posture
+
+
```
jimleroyer, line 88: We need disadvantages for removing SMS, with the most important con being the migration of users moving away from this authentication mechanism. Also, we need to remove phone numbers as required information part of the registration. We are losing one way to reach out to our users.

jimleroyer (commented):
This is a considerable move for our authentication mechanism, and hopefully the first ADR of many on that topic. My summarized thoughts:

1. We need to use the https://github.com/cds-snc/adr repository and not this one.
1. The ADR could be split in 2, as there are...
    1. valuable recommendations on future authentication modernization in GCNotify,
    1. but these are not tied to the initial stated problem: we need to make phone numbers not required through the registration process. More specifically, we do not need to drop SMS as an authentication mechanism to solve that issue, even though that would be favorable.
1. Hence the most pressing ADR might rather focus on the user registration workflow, and code refactoring to remove the requirement of having a phone number. A second ADR can be scoped to SMS removal and migration. A third one could be scoped to password-less authentication mechanisms.

records/2025-02-19.authmethods.md
```diff
@@ -0,0 +1,174 @@
...
+• Device loss can lead to account recovery challenges  
+• Slightly higher friction during initial setup  
+
+### Implementation:    
+The FIDO2 library that Notify currently uses for implementing YubiKey can be extended to include Authenticator apps.
+The implementation would follow the details for YubiKey and store login events in the sessions table.
+
+## 3. [Magic Link ADR](https://github.com/cds-snc/notification-adr/pull/64/files)
+We can revive the work done in the above ADR for magic links. There is a stock implementation in Notify, we could use for Magic Links.
+Please see the above ADR for introducing magic links into our 2FA flow
```
patheard, line 112: One comment I have about magic links is that we ran into issues with them on the ill fated URL Shortener project.  What we were seeing happen is that departmental email security features were testing the magic links and claiming the login session before the user could click the link.  

Our workaround was to keep the magic link valid for a set period of time rather that remove the magic link's hash once claimed (but this was less secure than what we had initially wanted).

records/2025-02-19.authmethods.md
```diff
@@ -0,0 +1,174 @@
...
+## Security Best Practices
+
+• Hardware security keys (like YubiKey) offer the strongest security   
+• Provide backup recovery methods   
+• Consider multiple authentication options for user flexibility   
+• If choosing between email and SMS, email generally offers better security against targeted attacks    
+
+# Recommendations
+
+The options below are not independent of each other. We can implement multiple options going forward.
```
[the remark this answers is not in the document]
jzbahrai: I meant to say independent of each other :/ I added the "not" by mistake

